PR #7, opened 2015-11-24
Update CF to write ELB logs to S3

danmassie:


jamespamplin:
:+1: looks good to me

PR #10, opened 2015-11-27
Add local certificates to gitignore.

susiecoleman:


mchv:
good idea to avoid any leak :+1:

PR #8, opened 2015-11-24
Styled up Sign in page

jamespamplin:
Implemented new sign in page styles:

signin-page-screen-shot-2015-11-24

- Not integrated, cannot sign in with email, yet (coming soon)
- Started using node-sass, replaced with [postcss](https://github.com/postcss/postcss) for greater extensibility with plugins
  - plugins currently integrated:
      - precss - sass like syntax for variables, mixins, and imports
      - cssnext - for transpiling future css syntax, autoprefixer for browser-prefixes, rem fallbacks
      - postcss-assets - for inlining and resolving asset urls in css
      - cssnano - for optimisation
- 3 Guardian webfonts inlined into CSS
- Inlined icons
- CSS styled checkbox for cross-browser consistency. Should fallback adequately
- Responsive layout based on core layout from [guardian/frontend](/guardian/frontend)
- Base 10px layout for nicer rem conversions

PR #11, opened 2015-11-30
 Signin endpoint

jamespamplin:
First pass at a endpoint to handle sign in form submissions.
- Identity API Client code isolated in `com.gu.identity.service.client` for potential extraction into own library later on
- Identity API Client uses abstractions for describing a http client and json parser. Should not be dependant on Play libraries.
- `com.gu.identity.frontend.services.IdentityService` used as Adapter from Identity API client code and Play APIs.

~~Based from #8, [compare here](https://github.com/guardian/identity-frontend/compare/postcss...signin-endpoint)~~ (Update, rebased from master now)


app/com/gu/identity/frontend/configuration/Configuration.scala
```diff
@@ -0,0 +1,24 @@
+package com.gu.identity.frontend.configuration
+
+import com.google.inject.ImplementedBy
```
mchv: why do we need dependency injection there?

app/com/gu/identity/frontend/controllers/SigninAction.scala
```diff
@@ -0,0 +1,63 @@
...
+    val rememberMe = getFormParam("keepMeSignedIn").contains("true")
+
+    identityService.authenticate(email, password, rememberMe).map {
+      case Left(errors) => redirectToSigninPageWithErrors(errors)
+      case Right(cookies) =>
+        SeeOther(getReturnUrl)
+          .withHeaders("Cache-Control" -> "no-cache")
+          .withCookies(cookies: _*)
+
+    }.recover {
```
mchv: In which exact cases do you need to recover?

app/com/gu/identity/frontend/controllers/SigninAction.scala
```diff
@@ -0,0 +1,63 @@
...
+      case NonFatal(ex) => {
+        logger.warn(s"Unexpected error signing in: ${ex.getMessage}", ex)
+        redirectToSigninPageWithErrors(Seq(ServiceGatewayError(ex.getMessage)))
+      }
+    }
+  }
+
+
+  private def getReturnUrl(implicit request: Request[AnyContent]) =
```
mchv: `request` does not need to be `implicit` there. `getReturnUrl(request)` is much more readable than  `getReturnUrl`

app/com/gu/identity/frontend/controllers/SigninAction.scala
```diff
@@ -0,0 +1,63 @@
...
+
+  def validateReturnUrl(returnUrl: String) =
+    returnUrl match {
+      case urlRegex(domain) if domain.endsWith(".theguardian.com") => true
+      case _ => false
+    }
+
+  private def redirectToSigninPageWithErrors(errors: Seq[ServiceError]) = {
+    val query = errors.map(e => s"error=${e.message}").mkString("&")
+    SeeOther(routes.Application.signIn().url + s"?$query")
```
mchv: `signIn` should accept an optional `Seq[ServiceError]` rather than construction the query yourself

app/com/gu/identity/frontend/services/IdentityService.scala
```diff
@@ -0,0 +1,45 @@
...
+import org.joda.time.{DateTime, Seconds}
+import play.api.mvc.{Cookie => PlayCookie}
+
+import scala.concurrent.{ExecutionContext, Future}
+
+
+/**
+ * Adapter for the identity service client.
+ */
+@ImplementedBy(classOf[IdentityServiceImpl])
```
mchv: idem about dependency injection

app/com/gu/identity/frontend/services/IdentityService.scala
```diff
@@ -0,0 +1,45 @@
...
+trait IdentityService {
+  def authenticate(email: Option[String], password: Option[String], rememberMe: Boolean)(implicit ec: ExecutionContext): Future[Either[Seq[ServiceError], Seq[PlayCookie]]]
```
mchv: Rather than exposing an execution context, it could make sense to have our own execution context for the identity service, but for an external service/module knowing which execution context to pass rather than the global one is difficult (they have no idea how you are going to use it).

app/com/gu/identity/service/client/ApiRequests.scala
```diff
@@ -0,0 +1,34 @@
...
+  val body: Option[String] = None
+}
+
+
+case class AuthenticateCookiesRequest(email: String, password: String) extends ApiRequest {
+  override val method = POST
+  override val headers = Seq("Content-Type" -> "application/x-www-form-urlencoded")
+  override val parameters = Seq("format" -> "cookies")
+  override val body = Some(s"email=$email&password=$password")
+}
```
mchv: Rather than constructing a new model of the HTTP request we should leverage the `WS` library (or any other) which will prevent us to have the need of an `HttpProvider`

app/com/gu/identity/service/client/Http.scala
```diff
@@ -0,0 +1,26 @@
...
+}
+
+
+trait HttpProvider {
+  def get(url: String, urlParameters: HttpParameters = Nil, headers: HttpParameters = Nil)(implicit ec: ExecutionContext): Future[Either[IdentityClientError, HttpResponse]]
+
+  def post(url: String, body: Option[String], urlParameters: HttpParameters = Nil, headers: HttpParameters = Nil)(implicit ec: ExecutionContext): Future[Either[IdentityClientError, HttpResponse]]
+
+  def delete(url: String, body: Option[String] = None, urlParameters: HttpParameters = Nil, headers: HttpParameters = Nil)(implicit ec: ExecutionContext): Future[Either[IdentityClientError, HttpResponse]]
+}
```
mchv: You should not do this  as there is already 2 implementations of the Http model in you dependencies (`playframework` and `WS` library)

app/com/gu/identity/service/client/Logging.scala
```diff
@@ -0,0 +1,7 @@
+package com.gu.identity.service.client
+
+import org.slf4j.LoggerFactory
+
+private[client] trait Logging {
+  protected val logger = LoggerFactory.getLogger(this.getClass)
+}
```
mchv: You should not need this, there is already a logger provided by the framework

app/com/gu/identity/service/client/package.scala
```diff
@@ -0,0 +1,26 @@
...
+  case class IdentityClientConfiguration(host: String, apiKey: String, httpProvider: HttpProvider, jsonParser: JsonParser)
+
+
+  trait JsonParser {
+    def extractAuthenticationCookiesResponse(body: String): Try[AuthenticationCookiesResponse]
+    def extractErrorResponse(body: String): Try[ApiErrorResponse]
+  }
+
+
+  case class Cookie(key: String, value: String, isSession: Boolean, expires: DateTime)
```
mchv: I think `Cookie` is already defined in the framework as well

app/com/gu/identity/service/client/ApiRequests.scala
```diff
@@ -0,0 +1,34 @@
...
+
+
+object AuthenticateCookiesRequest {
+  private val emailRegex = "^.+@.+$".r
```
mchv: You can reuse the [constraint validation framework](https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/scala/play/api/data/validation/Validation.scala#L77-L84) provided by play rather than implement you email regex.

app/com/gu/identity/service/client/IdentityClient.scala
```diff
@@ -0,0 +1,78 @@
...
+    AuthenticateCookiesRequest.from(email, password) match {
+      case Right(request) => authenticateCookies(request)
+      case Left(err) => Future.successful(Left(Seq(err)))
+    }
+
+  def authenticateCookies(request: AuthenticateCookiesRequest)(implicit configuration: IdentityClientConfiguration, ec: ExecutionContext): Future[Either[IdentityClientErrors, Seq[Cookie]]] =
+    makeRequest(request, ApiEndpoints.authenticateEndpoint).map {
+      case Left(err) => Left(err)
+      case Right(response) => {
+        configuration.jsonParser.extractAuthenticationCookiesResponse(response.body)
```
mchv: I am not sure why you need to parse json to extract the authentication cookie response

mchv:
> Identity API Client uses abstractions for describing a http client and json parser. Should not be dependant on Play libraries.
I think there is almost never a good reason to do this and you abstraction just add complexity to the implementation. Let's discuss tomorrow about this on how we could simply the code and keep the ability to switch to another `http` library without the need to redefine an `HTTP` model.

app/com/gu/identity/frontend/configuration/Configuration.scala
jamespamplin: Agreed with @mchv to look at Compile time DI separately: #12

app/com/gu/identity/frontend/controllers/SigninAction.scala
jamespamplin: Revised to use Play's forms to extract from post body. Could not use routes to extract from body.

app/com/gu/identity/frontend/controllers/SigninAction.scala
jamespamplin: Agreed. Will do this as part of the error messages work.

app/com/gu/identity/service/client/ApiRequests.scala
jamespamplin: Removed HTTP and JSON abstractions. Clients now use a `IdentityClientRequestHandler` with the following method:
```scala
handleRequest(request: ApiRequest): Future[Either[IdentityClientErrors, ApiResponse]] = ???
```
Enabling clients to implement and manage their own http logic.

app/com/gu/identity/service/client/Http.scala
jamespamplin: Removed.

app/com/gu/identity/service/client/Logging.scala
jamespamplin: Package level logger, for when client may be extracted to own library.

app/com/gu/identity/service/client/package.scala
jamespamplin: Revised this to a `IdentityCookie` which is a model object. The `IdentityCookie` represents the domain object of a cookie for a user. This is later transformed to a `PlayCookie` which represents an actual http cookie

app/com/gu/identity/service/client/ApiRequests.scala
jamespamplin: This is in the client so that all clients may reuse the same basic validation. Client is designed to have minimal dependencies and not dependant on Play.

app/com/gu/identity/service/client/IdentityClient.scala
jamespamplin: API returns json. Json abstraction has been replaced with a request handler abstraction.

jamespamplin:
Pushed up changes as discussed with @mchv. Hopefully a bit better now. We can iterate on suggestions in the future.

app/com/gu/identity/frontend/controllers/SigninAction.scala
```diff
@@ -0,0 +1,73 @@
...
+    val formParams = signInFormBody.bindFromRequest()(request).get
+
+    identityService.authenticate(formParams.email, formParams.password, formParams.rememberMe).map {
+      case Left(errors) => redirectToSigninPageWithErrors(errors)
+      case Right(cookies) =>
+        SeeOther(normaliseReturnUrl(formParams.returnUrl))
+          .withHeaders("Cache-Control" -> "no-cache")
+          .withCookies(cookies: _*)
+
+    }.recover {
```
mchv: `Action.async` returns a `Future`. When a future failed the framework will catch the exception and then call the [error handler](https://www.playframework.com/documentation/2.4.x/ScalaErrorHandling). So in theory you should not need to handle the recovering in the controller, and add the redirection to the sign in page in your custom error handler.

app/com/gu/identity/service/client/Errors.scala
```diff
@@ -0,0 +1,16 @@
...
+
+sealed trait IdentityClientError extends Throwable {
+  val message: String
+  val description: Option[String]
+}
+
+case class GatewayError(message: String,
+                        description: Option[String] = None,
+                        context: Option[String] = None,
+                        cause: Option[Throwable] = None) extends RuntimeException with IdentityClientError
```
mchv: Rather than `RuntimeException` the 2 errors should extend `scala.util.control.NoStackTrace` to avoid printing the stacktrace which is costly and useless.

mchv:
jamespamplin  :+1: I think the code is better. I added 2 small suggestions that could be addressed later in future PR.

conf/routes
```diff
@@ -6,6 +6,8 @@
 GET        /                              @com.gu.identity.frontend.controllers.Application.index
 GET        /signin                        @com.gu.identity.frontend.controllers.Application.signIn
 
+POST       /actions/signin                @com.gu.identity.frontend.controllers.SigninAction.signIn
```
danmassie: I wouldn't bother with /actions/* 

I'd put both the GET and POST /signin endpoints into the same Signin controller.

app/com/gu/identity/frontend/configuration/Configuration.scala
```diff
@@ -0,0 +1,24 @@
...
+}
+
+class ApplicationConfiguration extends Configuration {
+
+  private def getString(path: String) =
+    current.configuration.getString(path).getOrElse(sys.error(s"Missing configuration: $path"))
+
+  val identityCookieDomain = getString("identity.frontend.cookieDomain")
+
+  val identityApiKey = getString("identity.api.key")
```
danmassie: It would be nice to fail fast at boot if the configuration is invalid or has missing keys

jamespamplin: Eager binding on DI for Config perhaps? I'll add this to #12

jamespamplin: Eager binding on DI for Config perhaps? I'll add this to #12

app/com/gu/identity/frontend/controllers/SigninAction.scala
jamespamplin: Good idea, will create a ticket for global error handling.

conf/routes
jamespamplin: Good suggestion, will do this separately.

app/com/gu/identity/frontend/controllers/SigninAction.scala
```diff
@@ -0,0 +1,73 @@
...
+
+import scala.util.control.NonFatal
+
+
+/**
+ * Form actions controller
+ */
+class SigninAction @Inject() (identityService: IdentityService) extends Controller with Logging {
+
+  case class SignInRequest(email: Option[String], password: Option[String], rememberMe: Boolean, returnUrl: Option[String])
```
markjamesbutler: Please raise a separate ticket to add validation here.  Either ad-hoc or custom constraints (e.g. Password.)

app/com/gu/identity/frontend/services/IdentityService.scala
```diff
@@ -0,0 +1,45 @@
...
+}
+
+
+class IdentityServiceImpl @Inject() (config: Configuration, adapter: IdentityServiceRequestHandler) extends IdentityService {
+
+  implicit val clientConfiguration = IdentityClientConfiguration(config.identityApiHost, config.identityApiKey, adapter)
+
+  def authenticate(email: Option[String], password: Option[String], rememberMe: Boolean)(implicit ec: ExecutionContext) = {
```
markjamesbutler: You should use SignInRequest here, rather than single attributes.

app/com/gu/identity/frontend/services/IdentityServiceRequestHandler.scala
```diff
@@ -0,0 +1,87 @@
...
+    response.status >= 400 && response.status < 500
+
+  def handleErrorResponse(response: WSResponse): IdentityClientErrors =
+    response.json.asOpt[ApiErrorResponse]
+      .map(_.errors.map {
+        case e if isBadRequestError(response) => BadRequest(e.message, e.description, e.context)
+        case e => GatewayError(e.message, e.description, e.context)
+      })
+      .getOrElse {
+        logger.warn(s"Unexpected error response: ${response.status} ${response.statusText} ${response.body}")
```
markjamesbutler: This should be an error rather than warning.

app/com/gu/identity/service/client/ApiRequests.scala
```diff
@@ -0,0 +1,42 @@
...
+    Iterable("X-GU-ID-Client-Access-Token" -> s"Bearer ${configuration.apiKey}")
+
+  def apiEndpoint(path: String)(implicit configuration: IdentityClientConfiguration) =
+    s"https://${configuration.host}/$path"
+}
+
+case class AuthenticateCookiesRequest(url: String, email: String, password: String, extraHeaders: HttpParameters = Nil) extends ApiRequest {
+  override val method = POST
+  override val headers = Seq("Content-Type" -> "application/x-www-form-urlencoded") ++ extraHeaders
+  override val parameters = Seq("format" -> "cookies")
```
markjamesbutler: You should also send through: persistent=true for remember me requests.  This sets a flag within the cookie to inform the refresh cookie process that the SetCookie should have an expiry.

app/com/gu/identity/frontend/controllers/SigninAction.scala
```diff
@@ -0,0 +1,73 @@
...
+      "password" -> optional(text),
+      "rememberMe" -> default(boolean, false),
+      "returnUrl" -> optional(text)
+    )(SignInRequest.apply)(SignInRequest.unapply)
+  )
+
+
+  def signIn = Action.async { request =>
+
+    val formParams = signInFormBody.bindFromRequest()(request).get
```
markjamesbutler: We should fold here to catch errors rather than get.  See:

https://www.playframework.com/documentation/2.4.x/ScalaForms#Validating-a-form-in-an-Action

markjamesbutler:
:+1: Validation and persistent cookies alterations to be address in a separate PR.

jamespamplin:
Thanks guys. Will address comments separately.

PR #15, opened 2015-12-03
Add cache headers to responses

danmassie:


app/com/gu/identity/frontend/controllers/SigninAction.scala
```diff
@@ -30,14 +30,13 @@ class SigninAction @Inject() (identityService: IdentityService) extends Controll
 
 
   def signIn = Action.async { request =>
-
+  NoCache {
```
jamespamplin: Indentation has kind of broken here

jamespamplin:
Minor comment on indentation, but otherwise looks good :+1:

PR #17, opened 2015-12-04
Functional test signin

mario-galic:
@jamespamplin 

[Screencast](https://saucelabs.com/tests/44f97109113542b283485f4a148961b7)
[Test log](http://build.capi.gutools.co.uk:8111/viewLog.html?buildId=238&tab=buildResultsDiv&buildTypeId=Identity_FunctionalTests_IdentityFrontend)


functional-tests/project/build.properties
```diff
@@ -0,0 +1,4 @@
+#Activator-generated Properties
+#Thu Dec 03 09:52:49 GMT 2015
+template.uuid=e17acfbb-1ff5-41f5-b8cf-2c40be6a8340
```
jamespamplin: You can remove template and the comments above this

functional-tests/src/test/scala/SigninSpec.scala
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  feature("Sign in") {
+    scenario("User signs in with newly created Identity account") {
+      val testUser = new TestUser
+
+      When("I visit 'Sign in' page ")
+      val signin = new pages.Signin(testUser)
+      go.to(signin)
+      assert(signin.pageHasLoaded())
```
jamespamplin: Might be better to use matchers for nicer error messages, but this can be looked at separately.

functional-tests/src/test/scala/SigninSpec.scala
```diff
@@ -0,0 +1,78 @@
...
+
+      And("I submit the form to create my new Identity account")
+      register.submit()
+
+      Then("I should land on 'Registration Confirmation' page")
+      val registerConfirm = new RegisterConfirm
+      assert(registerConfirm.pageHasLoaded())
+
+      And("I should have Identity cookies.")
+      Seq("GU_U", "SC_GU_U", "SC_GU_LA").foreach { idCookie =>
```
jamespamplin: Cookies and Cookie checks could be abstracted out, but can be done separately.

functional-tests/src/test/scala/pages/Homepage.scala
```diff
@@ -0,0 +1,19 @@
+package test.pages
+
+import test.util.WebBrowserUtil
+import org.scalatest.selenium.{Page, WebBrowser}
+
+class Homepage extends Page with WebBrowser with WebBrowserUtil {
+  val url = "www.theguardian.com"
+
+  def userDisplayName: String = {
+    assert(pageHasElement(signIn))
```
jamespamplin: I don't think asserts should be done within a page object. Page objects should just be adapters between our code and selenium selectors for the page.

functional-tests/src/test/scala/util/WebBrowserUtil.scala
```diff
@@ -0,0 +1,54 @@
...
+import org.scalatest.selenium.WebBrowser
+import scala.util.{Success, Failure, Try}
+import scala.collection.JavaConverters._
+
+trait WebBrowserUtil { this: WebBrowser =>
+
+  lazy implicit val driver = Config.driver
+
+  protected def resetDriver() = {
+    driver.get("about:about")
```
jamespamplin: Why do we need to go to `about:about` on each reset?

functional-tests/src/test/scala/util/Config.scala
```diff
@@ -0,0 +1,53 @@
...
+      new RemoteWebDriver(url, capabilities)
+    }.getOrElse {
+      new ChromeDriver()
+    }
+  }
+
+  def webDriverSessionId(): String = {
+    Config.driver match {
+      case remoteDriver: RemoteWebDriver => remoteDriver.getSessionId.toString
+      case _ => throw new ClassCastException
```
jamespamplin: From the looks of this line, it just tries to get the session id for logging purposes. I think this could just fallback to a string like `"--unknown-id--"` or something similar, as an exception would fail a build.

jamespamplin:
:+1: this is a good start. Great work! Happy for feedback to be addressed in a separate PR.

PR #19, opened 2015-12-04
Move upstart into app artifact

danmassie:


frontend.sbt
```diff
@@ -23,3 +23,5 @@ sourceGenerators in Assets <+= build in Assets
 
 // Include handlebars views in resources for lookup on classpath
 unmanagedResourceDirectories in Compile += (resourceDirectory in Assets).value
+
+mappings in Universal ++= (baseDirectory.value / "deploy" * "*").get map (x => x -> ("deploy/" + x.getName))
```
jamespamplin: This should go in `build.sbt` rather than `frontend.sbt`. Separately, I'd like to create a `deploy.sbt` which has all the riffraff and other packaging config.

jamespamplin: Also, theres some builtin functions for pairing files for mappings, so I wonder if you can change this to something like:

```scala
mappings in Universal ++= (baseDirectory.value / "deploy" ***).get pair relativeTo(baseDirectory.value)
```
http://www.scala-sbt.org/0.13/docs/Mapping-Files.html

danmassie: I missed that there was a build.sbt! Different layout to the rest of the identity projects..

jamespamplin:
:+1: great stuff! Looks good.

PR #22, opened 2015-12-07
Set persistent parameter on auth request based on rememberMe

danmassie:


jamespamplin:
:+1: great stuff, thanks @danmassie

PR #23, opened 2015-12-07
Onminture analytics integration

jamespamplin:
First stab at integrating omniture analytics into the app.

- Ripped out omniture library and integration from frontend. Loads of code is commented out for reference just in case we might need it some day (eg: mvt stuff)
- Pack JS with [webpack](http://webpack.github.io/)
- Extended play's `SecurityHeadersFilter` so that we can dynamically declare `Content-Security-Policy` headers
  - Note: I'm not happy about how I've retrieved the hash for the inlined JS config, but we can iterate on this.
- Fixed the frontend assets build task to only run when client side stuff changes


app/com/gu/identity/frontend/controllers/Application.scala
```diff
@@ -14,7 +15,7 @@ class Application @Inject() extends Controller with Logging {
 
   def signIn(email: Option[String], error: Seq[String]) = Action { req =>
     Cached{
-      Ok(renderSignIn(error, email.getOrElse("")))
+      Ok(renderSignIn(configuration, error, email.getOrElse("")))
```
markjamesbutler: Minor point but it might be nicer to just pass through Omniture account details into ViewRender rather than entire config object.  Therefore decoupling ViewRenderer from the configuration.

app/com/gu/identity/frontend/views/models/InlineSource.scala
```diff
@@ -0,0 +1,22 @@
...
+case class InlineSource(source: String, sha256: String) extends ViewModel {
+  override def toMap = Map("source" -> source, "sha256" -> sha256)
+}
+
+object InlineSource {
+  def apply(source: String): InlineSource =
+    InlineSource(source, sha256(source))
+
+  private def sha256(in: String) =
+    Base64.encodeBase64String(DigestUtils.sha256(in))
```
markjamesbutler: This could return null if input is null.   Unlikely but might be worth wrapping in an Option.

markjamesbutler:
Couple of minor notes but LGTM :+1: Please create a ticket for compile time hash generation for Content-Security-Policy.

app/com/gu/identity/frontend/controllers/Application.scala
jamespamplin: Good thinking, although we could have a separate `JSConfiguration` case class which is a subset of `Configuration`. Will add this to tech debt

app/com/gu/identity/frontend/views/models/InlineSource.scala
jamespamplin: Done

PR #24, opened 2015-12-08
Add tracking data to id api requests.

danmassie:
Tracking Data and client ip extraction copied from frontend

### DIFF
--- a/app/com/gu/identity/frontend/models/TrackingData.scala
+++ b/app/com/gu/identity/frontend/models/TrackingData.scala
@@ -1,0 +1,34 @@
+package com.gu.identity.frontend.models
+
+import com.gu.identity.frontend.utils.RemoteAddress
+import com.gu.identity.service.client.HttpParameters
+import play.api.mvc.RequestHeader
+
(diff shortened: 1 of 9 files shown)


app/com/gu/identity/frontend/models/TrackingData.scala
```diff
@@ -0,0 +1,34 @@
...
+    registrationType.map("trackingRegistrationType" -> _),
+    omnitureSVi.map("trackingOmnitureSVI" -> _),
+    ipAddress.map("trackingIpAddress" -> _),
+    referrer.map("trackingReferer" -> _),
+    userAgent.map("trackingUserAgent" -> _)
+  ).flatten
+}
+
+object TrackingData extends RemoteAddress {
+  def apply(request: RequestHeader, returnUrl: Option[String]): TrackingData = {
```
jamespamplin: Not sure about modelling the tracking data in this way, but its fine for now. I think we should model these values in `SignInRequest` and extract them out in the client as needed. But we can revisit this later.

jamespamplin:
:+1: great work, thanks for this!

PR #26, opened 2015-12-09
UXD tweaks

jamespamplin:
sign-in-screen-shot-ux-tweaks-small

- 400px body
- drop the largest breakpoint
- footer padding to match membership
- buttons to be same height
- fix alignment of forgotten password and remember me
- fix remember me tick on iphones
- dropped bold font on heading
- revised horizontal rule colours
- fixed escaped quote in sign in conditions
- removed css compile warnings
- use flex-box to fill screen on large screens
- focus style on back button
- fixed remember me tick position in IE9


conf/messages.en
```diff
@@ -11,7 +11,7 @@ signin.rememberme=Remember me
 signin.signin=Sign in
 signin.noaccount=Don't have an account?
 signin.signup=Sign up
-signin.conditions=By proceeding, you agree to the Guardian's
```
markjamesbutler: Is this a typo?

public/components/fonts/_fonts.css
```diff
@@ -32,7 +32,10 @@
 
 /* GuardianTextEgyptianWeb */
 @mixin guFontFace GuardianTextEgyptianWeb, ascii, GuardianTextEgyptianWeb-Regular;
-@mixin guFontFace GuardianTextEgyptianWeb, ascii, GuardianTextEgyptianWeb-Bold, 700;
+
+/* Removed bold fonts for now as no longer used.
```
markjamesbutler: Lets remove this completely rather than commenting it out.  We can always find it in git history if required.

public/components/footer/_footer.css
```diff
@@ -14,7 +14,7 @@
   border-top: 1px solid #666;
```
markjamesbutler: Should this be using px or rem?

conf/messages.en
jamespamplin: Needed to escape the quote: https://playframework.com/documentation/2.4.x/ScalaI18N#Notes-on-apostrophes

public/components/footer/_footer.css
jamespamplin: Actually some things need to be pixels, especially when they're borders of 1 pixel size. If we use rems for 1px borders, they can sometimes get removed when the user is using small browser font settings, as the browser will not render borders < 1px of size. Will make a note of this when I write up the frontend-contributing stuff

public/components/fonts/_fonts.css
jamespamplin: Done.

public/components/footer/_footer.css
markjamesbutler: Cool lets add to the docs.  Its good to know for us without expert CSS know-how.

markjamesbutler:
LGTM :+1:

PR #27, opened 2015-12-09
Record MVT results in Omniture and Ophan

jamespamplin:
- Extracts user IDs from cookies as per original logic in frontend
- Record mvt results in Omniture and Ophan as per original logic in frontend


markjamesbutler:
LGTM :+1:

PR #28, opened 2015-12-10
Ab test cookie

danmassie:


app/com/gu/identity/frontend/filters/BetaUserGroupFilter.scala
```diff
@@ -0,0 +1,26 @@
...
+class BetaUserGroupFilter @Inject() (configuration: Configuration)  extends Filter {
+
+  private[filters] val CookieName = "GU_PROFILE_BETA"
+  private[filters] val CookieValue = "1"
+
+  override def apply(nextFilter: (RequestHeader) => Future[Result])(request: RequestHeader): Future[Result] = {
+    val cookie = Cookie(
+      name = CookieName,
+      value = CookieValue,
+      domain = Some(configuration.identityCookieDomain),
```
jamespamplin: Can we use a different property to restrict the cookie to just `profile.theguardian.com`, such as `profileCookieDomain` or something. Also Cookie can be secure only too.

jamespamplin:
Small note on Cookie domain, but otherwise looks good :+1:

app/com/gu/identity/frontend/filters/BetaUserGroupFilter.scala
jamespamplin: Actually just remembered that we also have `profile-beta.theguardian.com` - so for now I think we can keep this on `.theguardian.com`.

danmassie: It won't work on the "beta" site if its profile.theguardian.com

jamespamplin:
:+1: all good

PR #30, opened 2015-12-14
Add url params

susiecoleman:
Passing through skipConformation and return url values. Updating the links to social signins, forgot password and sign up so the skip confirmation and return url params are added to these links


app/com/gu/identity/frontend/views/models/SignInViewModel.scala
```diff
@@ -64,7 +84,30 @@ object SignInViewModel {
...
+      registerUrl = UrlBuilder("/register", urlParams),
+      forgotPasswordUrl = UrlBuilder("/reset", urlParams),
+      links = SignInLinksViewModel(urlParams)
     )
   }
 }
+
+object UrlBuilder {
+
+  def apply(baseUrl: String, params: Seq[Option[(String, String)]]) = {
```
jamespamplin: Is it possible to reduce `params` to the type `Seq[(String,String)`? As a `Seq(None)` should be the same as `Seq.empty`.

app/com/gu/identity/frontend/views/models/SignInViewModel.scala
```diff
@@ -64,7 +84,30 @@ object SignInViewModel {
...
+     createParamString(params) match {
+       case "" => baseUrl
+         //remove the final & added to the param list
+       case paramString => s"${baseUrl}/?${paramString}".dropRight(1)
```
jamespamplin: This could be simplified by using map over each param to create your key-value string, then use `mkString("&")` to join all strings with an `&` separator.

jamespamplin:
A couple of minor comments, but otherwise looks good :+1:

app/com/gu/identity/frontend/views/models/SignInViewModel.scala
susiecoleman: I don't think that would work as it will be a sequence of Somes and Nones so I don't think that makes it equivalent to Seq.empty

PR #31, opened 2015-12-14
Route assets on /static

jamespamplin:
Renamed `/assets` route to `/static` to avoid potential collisions with old app.


danmassie:
:+1:

PR #36, opened 2015-12-16
Ensure body params are URL encoded

jamespamplin:
Fixes issues with % signs in passwords.


markjamesbutler:
:+1:

emma-p:
Would it be worth adding a test?

jamespamplin:
Test added, thanks

PR #37, opened 2015-12-16
Enable Prout

mario-galic:
@jamespamplin 

* Added `/management/manifest` endpoint which Prout checks for commit SHA.
* Overdue deployment to CODE is set to 15 min limit
* Overdue deployment to PROD is set to 1 hour limit (since deploy to PROD is a manual process)

Tested on a [forked repository](https://github.com/mario-galic/identity-frontend/pull/2)


jamespamplin:
:+1: great work, thanks for this!

PR #38, opened 2015-12-16
Removed trailing slash from links

jamespamplin:
Fixes broken links on Prod for forgot password and sign up.


mchv:
:+1:  :boom:

PR #39, opened 2015-12-16
Tell Prout to disable SSL verification for CODE

rtyley:
...so that it can hit the manifest endpoint https://profile-origin.code.dev-theguardian.com/management/manifest - and doesn't say that https://github.com/guardian/identity-frontend/pull/38 is overdue.

image

See also https://github.com/guardian/prout/pull/16

cc @mario-galic


jamespamplin:
Ah, thanks @rtyley :+1:

PR #41, opened 2015-12-17
Remember email in sessionStorage

jamespamplin:
Remembers the user's email address in browser `sessionStorage` instead of using a URL parameter.

### DIFF
--- a/public/components/browser/storage.js
+++ b/public/components/browser/storage.js
@@ -1,0 +1,30 @@
+/*global window*/
+
+/**
+ * Basic abstraction around browser key/value storage, eg: sessionStorage and localStorage.
+ *
+ * @param {string} type
+ */
+function storage( type ) {
+  const storageActual = window[ type ];
+
+  function checkSupported() {
+    if ( !storageActual ) {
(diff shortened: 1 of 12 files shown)


jamespamplin:
Fixed merge conflicts due to  #44

public/components/browser/storage.js
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Basic abstraction around browser key/value storage, eg: sessionStorage and localStorage.
+ *
+ * @param {string} type
+ */
+function storage( type ) {
+  const storageActual = window[ type ];
+
+  function checkSupported() {
```
markjamesbutler: isSupported would read a little better.

markjamesbutler:
LGTM :+1:

public/components/browser/storage.js
jamespamplin: I've called this `checkSupported()` as it throws an error if it fails. `isSupported` would be ok if its just a boolean check.

Its unlikely for this code to throw as I have a high-level check for `sessionStorage` in `public/components/browser/browser.js`.

For now I think this is ok.

PR #44, opened 2015-12-22
Referer return url

markjamesbutler:
This PR implements redirecting the user back to the Referer when the returnUrl is not set by query parameter.  This is the same behaviour that exists on the existing dotcom::identity solution.


nlindblad:
:+1:

PR #53, opened 2016-01-12
MVT framework

jamespamplin:
Framework for defining MVT tests.

- Tests defined in [MultiVariantTests.scala](https://github.com/guardian/identity-frontend/blob/mvt-framework/app/com/gu/identity/frontend/configuration/MultiVariantTests.scala).
- Calculation of test participation ported from frontend.
- Docs for how to write a test in [CONTRIBUTING.md](https://github.com/guardian/identity-frontend/blob/mvt-framework/CONTRIBUTING.md#multi-variant-tests).
- Large changes to View models:
  - `Content-Security-Policy` now calculated dynamically from defined `Seq[PageResource]` on each `ViewModel with ViewModelResources` - at the moment, all resources are the same and defined on `BaseLayoutViewModel`
  - Removed necessity for `toMap` on `ViewModel`, and now using object of type `ViewModel with ViewModelResources with Product` for each view (aka case classes) instead of `Map[String, Any]`


CONTRIBUTING.md
```diff
@@ -109,6 +109,67 @@ Pixel fallbacks for `rem` units are added with PostCSS automatically via the
...
+access test results for client-side tests, use:
+
+```js
+import { getClientSideActiveTestResults } from 'components/analytics/mvt';
+
+const results = getClientSideActiveTestResults();
+```
+
+#### Recording test results
+Test results will be recorded on page view automatically in Omniture and Ophan.
```
mchv: Do we need to record the results in both `Ophan` and `Omniture`?
May we have the results only in `Ophan`?

app/com/gu/identity/frontend/configuration/MultiVariantTests.scala
```diff
@@ -0,0 +1,162 @@
...
+ *
+ * To define a test, create a {{case object}} which extends from
+ * {{MultiVariantTest}} and add it to {{MultiVariantTests.all}}.
+ *
+ * For example:
+ * {{{
+ * case object MyABTest extends MultiVariantTest {
+ *   val name = "MyAB"
+ *   val audience = 0.2
+ *   val audienceOffset = 0.6
```
emma-p: Maybe it'd be worth adding a validation to make sure that you cannot set incompatible `audience` and `audienceOffset` values, e.g. 0.2 & 0.9 ?

app/com/gu/identity/frontend/configuration/MultiVariantTests.scala
```diff
@@ -0,0 +1,162 @@
...
+  implicit val jsonWrites: Writes[MultiVariantTestVariant] = new Writes[MultiVariantTestVariant] {
+    override def writes(o: MultiVariantTestVariant): JsValue = Json.obj(
+      "id" -> o.id
+    )
+  }
+}
+
+
+
+case object SignInV2Test extends MultiVariantTest {
```
emma-p: Maybe this and the one below could be moved to separate files for more clarity

test/com/gu/identity/frontend/configuration/MultiVariantTestsSpec.scala
```diff
@@ -0,0 +1,95 @@
...
+import MultiVariantTests.{activeVariantForTest, isInTest}
+
+
+class MultiVariantTestsSpec extends WordSpec with Matchers {
+
+  "A Multi Variant Test" when {
+
+    "determining whether user is in test" should {
+
+      val test = RuntimeMultiVariantTest(
```
emma-p: Maybe a good case for using scalacheck? (Can be good to catch potential edge cases)

nlindblad: Agree :100:

app/com/gu/identity/frontend/views/models/ContentSecurityPolicy.scala
```diff
@@ -0,0 +1,52 @@
...
+
+  private def cspForResource(resource: PageResource): String = resource match {
+    case r: LocalResource => CSP_SELF_DOMAIN
+    case r: ScriptResource with InlinedSource => toCSPShaDefinition(r.sha256)
+    case r: InlinedResource => CSP_DATA_PROTOCOL
+    case r: ExternalResource => r.domain
+  }
+
+
+  def toCSPHeader(csp: Map[String, Seq[String]]) =
```
emma-p: Maybe add a test?

CONTRIBUTING.md
jamespamplin: As discussed, for now we need both Ophan and Omniture.

app/com/gu/identity/frontend/configuration/MultiVariantTests.scala
jamespamplin: Yes, agreed, can do this in the future. Will add a ticket.

test/com/gu/identity/frontend/configuration/MultiVariantTestsSpec.scala
jamespamplin: Yes, would like to do this, will do this in a separate ticket.

app/com/gu/identity/frontend/views/models/ContentSecurityPolicy.scala
jamespamplin: Basic tests added.

app/com/gu/identity/frontend/configuration/MultiVariantTests.scala
jamespamplin: For now, need to keep this in the same file as `MultiVariantTest` and `MultiVariantTestVariant` are sealed. But I've moved the JSON writes implicits, so file should be a little easier to read now.

emma-p:
:+1:

PR #56, opened 2016-01-14
Fix Return URL encoding for links

jamespamplin:
Ensures the Return URL and other parameters are encoded when appended to links in the views, such as the Social Sign in links.


susiecoleman:
:+1:

PR #59, opened 2016-01-18
Register endpoint 2nd pass

susiecoleman:
- Add more specific error messages
 - Handle skip confirmation
 - Handle 3rd party T&Cs
 - Input validation


app/com/gu/identity/frontend/configuration/Configuration.scala
```diff
@@ -23,5 +25,7 @@ class ApplicationConfiguration(val appConfiguration: PlayConfiguration) extends
   val identityApiKey = getString("identity.api.key")
   val identityApiHost = getString("identity.api.host")
 
+  val identityProfileBase = getString("identity.frontend.tandCsBase")
```
jamespamplin: Could this config key be named a bit simpler? ie: `identity.frontend.baseUrl`

conf/CODE.conf
```diff
@@ -4,6 +4,8 @@ identity.api.host=idapi.code.dev-theguardian.com
 
 identity.frontend.cookieDomain=code.dev-theguardian.com
 
+identity.frontend.base="https://profile.code.dev-theguardian.com"
```
jamespamplin: Doesn't appear to be used. Probably should be `identity.frontend.tandCsBase` or `identity.frontend.baseUrl` as I've suggested above.

app/com/gu/identity/frontend/controllers/RegisterAction.scala
```diff
@@ -41,21 +55,20 @@ class RegisterAction(identityService: IdentityService, val messagesApi: Messages
     NoCache {
       val clientIp = ClientRegistrationIp(request)
       registerForm.bindFromRequest.fold(
-        errorForm => Future.successful(SeeOther(routes.Application.register(Seq("error-registration"), None).url)),
+        errorForm => Future.successful(SeeOther(routes.Application.register(Seq("error-registration")).url)),
```
jamespamplin: Is there a way to surface which field has failed validation here? Will be useful to report this to the user.

app/com/gu/identity/frontend/configuration/Configuration.scala
susiecoleman: Changed

conf/CODE.conf
susiecoleman: fixed

app/com/gu/identity/frontend/controllers/RegisterAction.scala
```diff
@@ -63,4 +76,64 @@ class RegisterAction(identityService: IdentityService, val messagesApi: Messages
...
+
+  private def checkUserDataIsUnique(error: ServiceError): String = {
+    error.message match {
+      case "Username in use" => "register-error-username-in-use"
+      case "Email in use" => "register-error-email-in-use"
+      case _ => s"register-${error.id}"
+    }
+  }
+
+  private def redirectRegisterSuccess(cookies: Seq[PlayCookie], returnUrl: ReturnUrl, skipConfirmation: Option[Boolean], group: Option[String]) = {
```
jamespamplin: This function builds the redirect urls and the result, and could be separated into two functions. eg:

```scala
def registerSuccessResult( ... ) =
  SeeOther(registerSuccessRedirectUrl)
    .withCookies(cookies: _*)

def registerSuccessRedirectUrl( ... ) =
  ??? // logic to determine redirect url
```

app/com/gu/identity/frontend/controllers/RegisterAction.scala
```diff
@@ -63,4 +76,64 @@ class RegisterAction(identityService: IdentityService, val messagesApi: Messages
...
+    val groupCode = validateGroupCode(group)
```
jamespamplin: Is it possible to move `validateGroupCode` into the `registerForm` mapping?

app/com/gu/identity/frontend/views/models/ErrorViewModel.scala
```diff
@@ -11,9 +11,13 @@ object ErrorViewModel {
   }
 
   val errorMessages = Map(
-    "error-gateway" -> "There was a problem signing in; please try again.",
-    "error-bad-request" -> "Incorrect email or password; please try again.",
-    "error-registration" -> "Sorry an account could not be set up."
+    "signin-error-gateway" -> "There was a problem signing in; please try again.",
```
jamespamplin: Not important, but we should move these to the `messages.en` file. Can be addressed separately as tech debt.

app/com/gu/identity/frontend/utils/UrlDecoder.scala
```diff
@@ -0,0 +1,32 @@
+package com.gu.identity.frontend.utils
+
+object UrlDecoder {
+
+  def getQueryParams(url: String): Map[String, String] = {
```
jamespamplin: Not sure if we should be decoding URL params ourselves. Is there a library we can use instead? Or just use java's built in one? /cc @markjamesbutler @danmassie

app/com/gu/identity/frontend/controllers/RegisterAction.scala
susiecoleman: Added

app/com/gu/identity/frontend/controllers/RegisterAction.scala
susiecoleman: changed

app/com/gu/identity/frontend/controllers/RegisterAction.scala
susiecoleman: Yes, moved

app/com/gu/identity/frontend/views/models/ErrorViewModel.scala
jamespamplin: Agreed that we can address this as tech debt separately.

app/com/gu/identity/frontend/utils/UrlDecoder.scala
jamespamplin: Agreed that we can address this as tech debt.

jamespamplin:
:+1: great work!

app/com/gu/identity/frontend/utils/UrlDecoder.scala
susiecoleman: Moved into test directory as currently only used for testing.

PR #62, opened 2016-01-20
Locate by inner text due to FB randomization

mario-galic:
@jamespamplin 

Facebook is randomising ID attribute of `'Edit the info you provide'`. Sometimes the ID is u_0_q, and other times u_0_s. Thus it needs to be located by inner text.


jamespamplin:
:+1:

PR #57, opened 2016-01-14
Ensure frontend build process raises errors to sbt

jamespamplin:
Modifed frontend build task in sbt to properly throw errors when build task fails, ensures that builds fail on error.

Other enhancements:
- moved frontend build task to own sbt `AutoPlugin` for easier maintenance
- files can be ignored from build process (such as `hbs` as they're not used client-side yet), improves dev performance when modifying files
- extra build commands can be easily defined
- error output properly captured and will show in browser via Play

PR #63, opened 2016-01-20
Register page

jamespamplin:
Frontend for new Register page.

Built on top of #57, required before merge.

- Abstracted common form styles to `public/components/form-foundations`, shared across sign in and register forms
- Social sign in, now called `oauth-cta` component - created own view models and extracted host to config, shared across sign in and register forms
- Terms (and conditions) extracted to own component
- Revised text to use modelled objects in `com.gu.identity.frontend.models.text._`

identity-frontend-register


app/com/gu/identity/frontend/models/Text.scala
```diff
@@ -35,15 +35,6 @@ object Text {
...
-  object SocialSignInText {
-    def toMap(implicit messages: Messages): Map[String, String] = {
-      Map(
-        "facebook" -> messages("socialsignin.description", "Facebook"),
-        "google" -> messages("socialsignin.description", "Google")
-      )
-    }
-  }
-
   object HeaderText {
```
susiecoleman: Should this file be moved into the text directory?

public/components/register-form/_register-form.hbs
```diff
@@ -0,0 +1,55 @@
...
+    </div>
+
+    <div class="register-form__control--username">
+        <label class="register-form__label--username" for="register_field_username">{{ registerPageText.username }}</label>
+        <input class="register-form__field--username" id="register_field_username" type="text" name="username" placeholder="{{ registerPageText.usernameHelp }}" autocomplete="on" autocapitalize="off" autocorrect="off" spellcheck="false" required minlength="6" maxlength="20" pattern="[a-zA-Z0-9]+" />
+    </div>
+
+    <div class="register-form__control--password">
+      <label class="register-form__label--password" for="register_field_password">{{ registerPageText.password }}</label>
+      <input class="register-form__field--password" id="register_field_password" type="password" name="password" placeholder="{{ registerPageText.passwordHelp }}" required minlength="6" />
```
susiecoleman: Password field also has a max length of 20

app/com/gu/identity/frontend/models/Text.scala
jamespamplin: Yes-ish, but I'd like to break this up into files per text object, and create type classes for each text object. As per text classes for register in `com.gu.identity.frontend.models.text.RegisterText`. I've created a ticket to do this as tech debt, already too much going in this PR.

public/components/register-form/_register-form.hbs
jamespamplin: I've purposely left off max-length as the built in html validation for this will prevent the user from typing when they reach the max-length, which may result in an unknown password if you are not watching the input when you type. We should allow any length password here, and produce an error with Javascript when the user exceeds the limit. I'll create a ticket to do this separately.

Although I'd prefer having a much higher limit than 20 for password length generally.

susiecoleman:
Looks good to me, @danmassie would you be able to have a closer look at the frontend.sbt file and frontendBuildPlugin.scala files if possible as they were both a bit beyond my understanding. Thanks

jamespamplin:
@susiecoleman actually those sbt things are in #57 which this PR is based from. /cc @danmassie

susiecoleman:
Awesome, in that case LGTM :+1:

PR #64, opened 2016-01-21
Add email validation success page

susiecoleman:
completeregistration

Implementing the page confirming the user has successfully registered.


public/register-confirmation-page.hbs
```diff
@@ -0,0 +1,15 @@
+{{# partial "content" }}
+
+  <div class="register-confirm">
+    <h1 class="register-confirm__heading">Now check your email</h1>
```
jamespamplin: Can we please move the text into `messages.en`

markjamesbutler:
@susiecoleman Could you please add a short description.  This helps others to understand the feature this PR is implementing.

public/register-confirmation-page.hbs
susiecoleman: Fixed

app/com/gu/identity/frontend/models/Text.scala
```diff
@@ -66,4 +66,18 @@ object Text {
       )
     }
   }
+
```
susiecoleman: This object needs to be moved out to it's own file to match the structure of the text in #63

jamespamplin: Up to you whether you do this in this PR, or we can address it as tech debt. I already have a card for messages API tech debt.

public/register-confirmation-page.hbs
```diff
@@ -0,0 +1,17 @@
+{{# partial "content" }}
```
jamespamplin: `pageTitle` partial needs to be set too, see: https://github.com/guardian/identity-frontend/blob/master/public/register-page.hbs#L1

susiecoleman: Added

jamespamplin:
:+1: great work!

PR #69, opened 2016-01-22
Long life caching of Assets

jamespamplin:
Use [`sbt-digest`](https://github.com/sbt/sbt-digest) to create hashes of assets on package compilation, which Play can use at runtime to cache the asset for a looonnnggg time.

Note: only works when app is running in production mode.


danmassie:
:+1:

PR #70, opened 2016-01-25
CSRF validation

jamespamplin:
Uses adapters around play's CSRF filters / API to use request tokens to prevent CSRF attacks.

Added to both Sign in and Registration endpoints.

### DIFF
--- a/app/com/gu/identity/frontend/csrf/CSRFActions.scala
+++ b/app/com/gu/identity/frontend/csrf/CSRFActions.scala
@@ -3,7 +3,7 @@ package com.gu.identity.frontend.csrf
 import com.gu.identity.frontend.controllers.NoCache
 import play.api.mvc._
 import play.filters.csrf.CSRF.ErrorHandler
-import play.filters.csrf.{CSRFAddToken => PlayCSRFAddToken, CSRFCheck => PlayCSRFCheck, CSRF, CSRFConfig}
+import play.filters.csrf.{CSRFAddToken => PlayCSRFAddToken, CSRFCheck => PlayCSRFCheck, CSRF}
 
 import scala.concurrent.Future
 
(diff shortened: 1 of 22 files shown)


jamespamplin:
Double checked that this CSRF logic does not break badly if the form is loaded with a bad csrf token (as it does with the current edit profile pages). The forms will still load, and when the user submits with a bad token, they are returned to the form with an error message, and a correct form, allowing them to resubmit.

Will create an issue in playframework to fix this issue properly.

app/com/gu/identity/frontend/csrf/CSRFActions.scala
```diff
@@ -0,0 +1,40 @@
...
+import play.api.mvc._
+import play.filters.csrf.CSRF.ErrorHandler
+import play.filters.csrf.{CSRFAddToken => PlayCSRFAddToken, CSRFCheck => PlayCSRFCheck, CSRF}
+
+import scala.concurrent.Future
+
+
+sealed trait CSRFActions
+
+case class CSRFAddToken(config: CSRFConfig) extends ActionBuilder[Request] with CSRFActions {
```
emma-p: Would it be worth adding a test here for CSRFActions? (seems to be the core of the logic)

emma-p:
:+1: Looks good to me apart from the comment

app/com/gu/identity/frontend/csrf/CSRFActions.scala
jamespamplin: Added `CSRFActionsSpec` to do some basic checks around the action compositions without doing a full integration test.

For the tests, I needed to create `com.gu.identity.frontend.test.SimpleFakeApplication` which creates a Fake app for tests as `play.api.test.FakeApplication` doesn't work with compile time (manual) DI applications. Some of the CSRF logic in play requires access to the Play config globals.

emma-p:
:+1:

PR #71, opened 2016-01-26
Removed beta user group filter

jamespamplin:
Removed the `BetaUserGroupFilter` as its function was to generate the `GU_PROFILE_BETA` cookie, which from guardian/identity-platform#109 will now be generated on edge in Fastly.

Cookie is only used for routing in Fastly.


markjamesbutler:
:shipit: :+1:

PR #74, opened 2016-01-27
Add recaptcha

susiecoleman:
Add google recaptcha to sign in page. Recaptcha is only displayed if the user has already failed to sign in once. They get one go without the recaptcha.
recaptchaonsignin


app/com/gu/identity/frontend/controllers/SigninAction.scala
```diff
@@ -36,22 +40,42 @@ class SigninAction(identityService: IdentityService, val messagesApi: MessagesAp
       val trackingData = TrackingData(request, formParams.returnUrl)
       val returnUrl = ReturnUrl(formParams.returnUrl, request.headers.get("Referer"))
 
-      identityService.authenticate(formParams.email, formParams.password, formParams.rememberMe, trackingData).map {
-        case Left(errors) => redirectToSigninPageWithErrorsAndEmail(errors, returnUrl, formParams.skipConfirmation)
-        case Right(cookies) => {
-          SeeOther(returnUrl.url)
-            .withCookies(cookies: _*)
+      formParams.googleRecaptchaResponse match {
```
jamespamplin: We could separate the recaptcha logic to a Action Composition or as a separate wrapper function like `NoCache`. This will separate its logic from that of the sign-in, and will also allow us to reuse the logic elsewhere.

app/com/gu/identity/frontend/controllers/SigninAction.scala
```diff
@@ -26,7 +29,8 @@ class SigninAction(identityService: IdentityService, val messagesApi: MessagesAp
       "password" -> optional(text),
       "rememberMe" -> default(boolean, false),
       "returnUrl" -> optional(text),
-      "skipConfirmation" -> optional(boolean)
+      "skipConfirmation" -> optional(boolean),
+      "g-recaptcha-response" -> optional(text)
```
jamespamplin: I think we should rename this field from `xxx-response` as response may be confusing with a http response. What is stored in this field? Is it a token? If so, perhaps `g-recaptcha-token`?

app/com/gu/identity/frontend/services/GoogleRecaptchaServiceHandler.scala
```diff
@@ -0,0 +1,50 @@
...
+    }
+  }
+
+  def getRecaptchaResponseFromGoogle(captchaResponseCode: String): Future[GoogleResponse] = {
+    ws.url("https://www.google.com/recaptcha/api/siteverify").post(
+      Map("secret" -> Seq(configuration.googleRecaptchaSecretKey), "response" -> Seq(captchaResponseCode))
+    ).map{
+      googleResponse => {
+        val body = googleResponse.body
+        Json.parse(body).as[GoogleResponse]
```
jamespamplin: `WSResponse` has a `json` method on it which can be used to extract a json response. `asOpt` might be better too for handling of errors for when the response cannot be parsed.

app/com/gu/identity/frontend/views/models/SignInViewModel.scala
```diff
@@ -31,8 +34,30 @@ case class SignInViewModel private(
...
+      createSignInViewModel(layout, configuration, returnUrl, skipConfirmation, errors, resources)
+    } else {
+      val recaptchaResources = GoogleRecaptchaResources(configuration)
+      val resources = layout.resources ++ recaptchaResources.resources
+      val recaptchaSiteKey = Some(recaptchaResources.googleSiteKey)
+      createSignInViewModel(layout, configuration, returnUrl, skipConfirmation, errors, resources, recaptchaEnabled = true, recaptchaSiteKey)
+    }
+  }
+
+  private def createSignInViewModel(
```
jamespamplin: The `apply` method should be responsible for creating a `SignInViewModel`. `GoogleRecaptchaResources` could be reworked to `GoogleRecaptchaViewModel` storing everything about a Google Recaptcha. Then `SignInViewModel` can store a `recaptcha: Option[GoogleRecaptchaViewModel]` which is a `Some(...)` when there are errors, and recaptcha is enabled.

conf/CODE.conf
```diff
@@ -10,5 +10,7 @@ identity.federation.api.host="https://oauth.code.dev-theguardian.com"
 
 omniture.account="guardiangudev-code"
 
+google.recaptchaIsDisabled = false
```
jamespamplin: Code is inconsistent where some is `recaptchaIsDisabled` and the config is `recaptchaEnabled`. I prefer always using positive flag settings for consistency, so `google.recaptchaEnabled = true`.

app/com/gu/identity/frontend/controllers/SigninAction.scala
susiecoleman: This is a parameter automatically added to the post request by the google recaptcha script
https://developers.google.com/recaptcha/docs/verify

jamespamplin: Okey dokey, best leave it as is then.

jamespamplin:
Final note, style wise can we please add a bit of a margin 5-10px above the captcha. Can we make it full width? if not perhaps centering it will help.  @chriswilk may have an opinion on this too.

conf/CODE.conf
mchv: :+1: to what james wrote

susiecoleman: fixed

app/com/gu/identity/frontend/views/models/SignInViewModel.scala
susiecoleman: Made some changes to this logic

app/com/gu/identity/frontend/controllers/SigninAction.scala
susiecoleman: Changed

app/com/gu/identity/frontend/services/GoogleRecaptchaServiceHandler.scala
susiecoleman: Changed

susiecoleman:
Updated recaptcha css
signin-recaptcha

app/com/gu/identity/frontend/views/models/ContentSecurityPolicy.scala
```diff
@@ -8,9 +8,11 @@ object ContentSecurityPolicy {
   val CSP_STYLE_SRC = "style-src"
   val CSP_FONT_SRC = "font-src"
   val CSP_IMG_SRC = "img-src"
+  val CSP_CHILD_SRC = "frame-src"
```
jamespamplin: Value should be called `CSP_FRAME_SRC` - we'll deal with `child-src` in the future.

jamespamplin:
:+1: great work!

PR #75, opened 2016-01-28
Fix issue where empty group code is in redirect URL on errors

jamespamplin:
Fixes an issue on PROD / CODE where the group code is on the redirect when an error occurs. This causes fastly to route to the old sign in form which doesn't show the error.


emma-p:
:+1:

PR #77, opened 2016-01-29
Remember form inputs on Register page

jamespamplin:
Remember form inputs on Register page using session storage in the browser as persistence.

- Uses classes to model the Register form, fields, and state
- Updated logic for sign in field persistence to match Register's
- Added lodash as a dependency, only using `mapValues` for now
- Changed webpack config to not mangle JS variables with Uglify, aids with debugging


susiecoleman:
I've had a look through this and it looks good. But I'm not clear on all of it, I'm happy to look through it together if you'd like?

susiecoleman:
:+1:

PR #78, opened 2016-02-02
Report csp violations

susiecoleman:
When The CSP is violated a POST request is made to csp/report and the violation is logged

### DIFF
--- a/app/com/gu/identity/frontend/views/models/ContentSecurityPolicy.scala
+++ b/app/com/gu/identity/frontend/views/models/ContentSecurityPolicy.scala
@@ -1,5 +1,6 @@
 package com.gu.identity.frontend.views.models
 
+import com.gu.identity.frontend.controllers._
 
 object ContentSecurityPolicy {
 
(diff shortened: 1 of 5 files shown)


app/com/gu/identity/frontend/views/models/ContentSecurityPolicy.scala
```diff
@@ -14,10 +14,17 @@ object ContentSecurityPolicy {
   val CSP_DATA_PROTOCOL = "data:"
   val CSP_UNSAFE_INLINE = "'unsafe-inline'"
 
+  val CSP_REPORT_URI = "report-uri"
+  val CSP_REPORT_URI_PATH = "/csp/report"
```
jamespamplin: Can we look this up via the router?

susiecoleman: Changed

jamespamplin:
:+1: looks good to me

PR #79, opened 2016-02-02
Error pages

jamespamplin:
Global error handler to display nicer error pages to the user.

Also revised `com.gu.identity.frontend.configuration.Configuration` to be a case class with a `testConfiguration` val for simpler testing.

identity-frontend-500

identity-frontend-404


jamespamplin:
Cleaned up test as per feedback from @emma-p and @susiecoleman

emma-p:
:+1:

PR #80, opened 2016-02-03
Better Content-Security-Policy 1.0 support

jamespamplin:
Found an issue related to CSP and Firefox version < 43, where only CSP 1.0 is properly supported. Hashes of inlined scripts are not supported in this browser resulting in client errors.

This PR changes the config from an inlined script to be statically injected into the response into a `<script type="application/json">` tag, then parsed via the Javascript when it initialises.


susiecoleman:
Talked through with @jamespamplin 
:+1:

PR #81, opened 2016-02-03
Move CSP report endpoint

jamespamplin:
Moved Content Security Policy endpoint under /actions to avoid having to have another route whitelisted in the fastly routing config.


susiecoleman:
:+1:

PR #82, opened 2016-02-03
Fixed broken CSP test due to #81

jamespamplin:
- Fixed broken tests for CSP due to changing the route for the report in
  #81


susiecoleman:
:+1:

PR #83, opened 2016-02-03
Ensure test results recorded for 100% tests

jamespamplin:
Fixes an issue on PROD where test results may be omitted when a `GU_mvt_id` cookie is not present by defining default variants on tests with 100% participation.

Also addresses some tech debt raised in #53:
- Moved all MVT stuff to `mvt` package
- Moved a lot of code out of `MultiVariantTests.scala` so that now it only contains Test definitions
- Created `ActiveMultiVariantTests` type definition for active test results instead of `Iterable[(MultiVariantTest, MultiVariantTestVariant)]` for better readability


nlindblad:
:+1:

PR #87, opened 2016-02-04
Return url tech debt

susiecoleman:
Addresses 3 tech debt issues with return url:

- If the return url is the default return url (i.e. http://www.theguardian.com) it is not included as a query param in the url to make the url cleaner

- Ignore return url and referrer header if they are the url for the sign in page, register page or register confirmation page.

- The default return url should be environment specific

### DIFF
--- a/app/com/gu/identity/frontend/models/ReturnUrl.scala
+++ b/app/com/gu/identity/frontend/models/ReturnUrl.scala
@@ -1,17 +1,21 @@
 package com.gu.identity.frontend.models
 
 import java.net.URI
+import com.gu.identity.frontend.configuration.Configuration
+
 import scala.util.Try
 
 case class ReturnUrl(url: String)
 
 object ReturnUrl {
 
-  val default = ReturnUrl("http://www.theguardian.com")
   val domains = List(".theguardian.com", ".code.dev-theguardian.com", ".thegulocal.com")
   val invalidUrlPaths = List("/signin", "/register", "/register/confirm")
 
-  def apply(returnUrl: Option[String], referer: Option[String]): ReturnUrl = {
+  def apply(returnUrl: Option[String], referer: Option[String], configuration: Configuration): ReturnUrl = {
+
+    val default = ReturnUrl(configuration.identityDefaultReturnUrl)
+
     returnUrl.map(ReturnUrl(_))
       .orElse(referer.map(ReturnUrl(_)))
       .filter(validDomain(_))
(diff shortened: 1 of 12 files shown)


app/com/gu/identity/frontend/models/ReturnUrl.scala
```diff
@@ -22,6 +28,12 @@ object ReturnUrl {
...
   }
 
+  def validUrlPath(returnUrl: ReturnUrl): Boolean = {
+    val urlPath = path(returnUrl)
+    !invalidUrlPaths.contains(urlPath)
+  }
+
   def host(returnUrl: ReturnUrl): String = Try(new URI(returnUrl.url)).map(_.getHost).getOrElse("")
 
+  def path(returnUrl: ReturnUrl): String = Try(new URI(returnUrl.url)).map(_.getPath).getOrElse("")
```
jamespamplin: Since we're parsing the `url` as a `String` into a `URI` twice now, perhaps it makes sense to store the parsed `URI` on the `ReturnUrl` object instead of a string. ie:

```scala
case class ReturnUrl(uri: URI) {
    lazy val url: String = uri.toString
}
```

This ensures we have a correct representation of a URI stored within the object, and we only parse it once.

test/com/gu/identity/frontend/models/ReturnUrlSpec.scala
```diff
@@ -17,19 +20,34 @@ class ReturnUrlSpec extends FlatSpec with Matchers {
...
+    ReturnUrl(None, None, config) should be(ReturnUrl("http://www.theguardian.com"))
 
-    ReturnUrl(None, Some("http://www.thegulocal.com/")) should be(ReturnUrl("http://www.thegulocal.com/"))
-    ReturnUrl(None, None) should be(ReturnUrl("http://www.theguardian.com"))
+    ReturnUrl(Some("http://baddomain.com"), None, config) should be(ReturnUrl("http://www.theguardian.com"))
+    ReturnUrl(None, Some("http://baddomain.com"), config) should be(ReturnUrl("http://www.theguardian.com"))
 
-    ReturnUrl(Some("http://baddomain.com"), None) should be(ReturnUrl("http://www.theguardian.com"))
-    ReturnUrl(None, Some("http://baddomain.com")) should be(ReturnUrl("http://www.theguardian.com"))
+    ReturnUrl(None, Some("http://profile-origin2.thegulocal.com"), config) should be(ReturnUrl("http://profile-origin2.thegulocal.com"))
```
jamespamplin: Can we have a test for something which isn't a real URL. ie: not parsable

app/com/gu/identity/frontend/models/ReturnUrl.scala
susiecoleman: Changed structure of code so that url string is not being parsed multiple times

jamespamplin:
:+1: looks good to me

PR #88, opened 2016-02-09
Fix issue where asset build triggers twice

jamespamplin:
- Fixes issue where both build-css and build-js commands run whenever a
  css file or a js file are modified. Caused by the CSS task incorrectly
  including the built JS source map as its task output, causing its
  incremental cache to invalidate.
- Also added debug logs for easier debugging of task outputs. Use `last`
  in sbt to see debug logs.


markjamesbutler:
:+1:

PR #89, opened 2016-02-10
Fix missing fields when Registering

jamespamplin:
Fixes the following missing fields when registering new users:

- Surname should be `secondName` as required by the API
- `receiveGnmMarketing` should be in a `statusFields` object
- `receive3rdPartyMarketing` should be in a `statusFields` object


nlindblad:
:+1:

PR #90, opened 2016-02-11
Update README to reflect S3 policies

nlindblad:
We've made it possible to grant other team accounts access to the S3 objects mentioned in the README.

This update provides more information about that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ aws --profile identity s3 cp s3://identity-local-ssl/profile-origin-thegulocal-c
 aws --profile identity s3 cp s3://identity-local-ssl/profile-origin-thegulocal-com-exp2016-11-10.key ./nginxCerts 1>/dev/null
 ```
 
-**Note**: If you do not have Janus access to Identity, we can grant your team specific access, which means you would substitute `--profile identity` with e.g. `--profile membership`.
+**Note**: If you do not have Janus access to Identity, we can grant your team specific access, which means you would substitute `--profile identity` with e.g. `--profile membership`. Contact the Identity team if you require access to these files.
 
 * Find the configuration folder of nginx by running:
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -71,9 +73,12 @@ You should now be able to start the application (`sbt run`), go to [https://prof
 Install the local configuration file from s3:
 
 ```
+mkdir -p /etc/gu
 aws s3 cp --profile identity s3://gu-identity-frontend-private/DEV/identity-frontend.conf /etc/gu
 ```
 
+**Note**: If you do not have Janus access to Identity, we can grant your team specific access, which means you would substitute `--profile identity` with e.g. `--profile membership`.
```
markjamesbutler: "Contact the Identity team if you require access to these files."

README.md
```diff
@@ -10,7 +10,7 @@ Web frontend for Sign in and Registration at [theguardian.com](http://theguardia
 Configuration files:
 - Environment-specific configuration (`conf/<ENV>.conf`)
 - Application configuration (`conf/application.conf`)
-- System file with additional properties (`/etc/gu/identity-frontend.conf`)
+- System file with additional properties (`/etc/gu/identity-frontend.conf` or `~/.gu/identity-frontend.conf`)
```
jamespamplin: I think we only read from `/etc/gu/identity-frontend.conf`, see: https://github.com/guardian/identity-frontend/blob/master/conf/DEV.conf#L18

But would be good to read from `~/.gu/` (in fact I actually map my `~/.gu/` to `/etc/gu`, so would be nice not to have to do this)

nlindblad: Okey. I'll look into reading from `~/.gu` as well. Cheers.

nlindblad: Added `~/.gu`

jamespamplin:
:+1: looks good

PR #91, opened 2016-02-19
Log registration and signin

susiecoleman:
Sends a custom metric to Cloudwatch whenever a user logs in or registers successfully.

Tested on CODE


app/com/gu/identity/frontend/configuration/Configuration.scala
```diff
@@ -82,4 +87,22 @@ object Configuration {
...
 
+  object AWSConfig {
+    val credentials: AWSCredentialsProvider = {
+      val provider = new AWSCredentialsProviderChain(
+        new EnvironmentVariableCredentialsProvider(),
+        new SystemPropertiesCredentialsProvider(),
+        new ProfileCredentialsProvider(),
+        new InstanceProfileCredentialsProvider
+      )
+      provider.getCredentials
```
jamespamplin: What does `provide.getCredentials` do here?

Also, can we please add an additional profile credentials provider to the chain for the `identity` profile, very helpful for local dev:

```scala
new ProfileCredentialsProvider("identity")
```

app/com/gu/identity/frontend/logging/MetricsActor.scala
```diff
@@ -0,0 +1,26 @@
+package com.gu.identity.frontend.logging
+import akka.actor._
+import play.api.Logger
+
+trait MetricsLoggingActor {
+  private val system = ActorSystem()
+  val successfulActionLogger = system.actorOf(Props[MetricsActor])
```
jamespamplin: I think this will create a new actor instance of `MetricsActor` each time `MetricsLoggingActor` is mixed into another class, but we probably only need a single instance. Suggest that we pull the actor out into its own `object` and instantiate the actor there.

app/com/gu/identity/frontend/controllers/RegisterAction.scala
```diff
@@ -129,6 +129,7 @@ class RegisterAction(identityService: IdentityService, val messagesApi: Messages
   }
 
   private def registerSuccessResult(url: String, cookies: Seq[PlayCookie]) = {
+    successfulActionLogger ! Register
```
jamespamplin: Since this message is static, suggest that you wrap this in a function within `MetricsLoggingActor`, allowing you to keep the Messages private to `MetricsActor.scala`, eg something like:
```scala
def logSuccessfulRegister(): Unit =
  successfulActionLogger ! Register
```

app/com/gu/identity/frontend/logging/SuccessfulActionCloudwatchLogging.scala
```diff
@@ -0,0 +1,56 @@
+package com.gu.identity.frontend.logging
+
+import com.amazonaws.handlers.AsyncHandler
+import com.amazonaws.regions.{Regions, Region}
+import com.amazonaws.services.cloudwatch.AmazonCloudWatchAsyncClient
+import com.amazonaws.services.cloudwatch.model.{Dimension, MetricDatum, PutMetricDataRequest}
+import com.gu.identity.frontend.configuration.Configuration._
+
+trait LoggingAsyncHandler extends AsyncHandler[PutMetricDataRequest, Void] with Logging {
```
jamespamplin: Does this need to be a trait?

app/com/gu/identity/frontend/logging/SuccessfulActionCloudwatchLogging.scala
```diff
@@ -0,0 +1,56 @@
...
+  }
+}
+
+object LoggingAsyncHandler extends LoggingAsyncHandler
+
+object SuccessfulActionCloudwatchLogging {
+
+  private lazy val cloudwatch = {
+    val client = new AmazonCloudWatchAsyncClient(AWSConfig.credentials, AWSConfig.clientConfiguration)
+    client.setEndpoint(Region.getRegion(Regions.EU_WEST_1).getServiceEndpoint(com.amazonaws.regions.ServiceAbbreviations.CloudWatch))
```
jamespamplin: Can we pull Region out into `AWSConfig`?

Plus, we can derive the Region from the instance using [`com.amazonaws.regions.Regions.getCurrentRegion`](http://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/regions/Regions.html#getCurrentRegion%28%29), and fallback to `EU_WEST_1` for dev environments.

app/com/gu/identity/frontend/logging/SuccessfulActionCloudwatchLogging.scala
```diff
@@ -0,0 +1,56 @@
...
+    val request = new PutMetricDataRequest()
+      .withNamespace("SuccessfulRegistrations")
+      .withMetricData(
+        new MetricDatum()
+          .withMetricName("SuccessfulRegistration")
+          .withUnit("Count")
+          .withValue(1d)
+          .withDimensions(stageDimension)
+      )
+    cloudwatch.putMetricDataAsync(request, LoggingAsyncHandler)
```
jamespamplin: Bit of duplication here, can we put this logic into a function to be a bit more DRY?

app/com/gu/identity/frontend/controllers/RegisterAction.scala
susiecoleman: Fixed for register and signin

app/com/gu/identity/frontend/configuration/Configuration.scala
susiecoleman: Is it best to leave the provide.getCredentials so it errors early or let it error somewhere else?

app/com/gu/identity/frontend/logging/MetricsActor.scala
susiecoleman: fixed

app/com/gu/identity/frontend/logging/SuccessfulActionCloudwatchLogging.scala
susiecoleman: Changed to object

app/com/gu/identity/frontend/logging/SuccessfulActionCloudwatchLogging.scala
susiecoleman: changed

app/com/gu/identity/frontend/configuration/Configuration.scala
jamespamplin: I don't think its necessary now, as `AWSConfig` is defined as an object, which is evaluated lazily, so the credentials will only get validated when used for the first time. So we can remove the extra call to `provider.getCredentials`

app/com/gu/identity/frontend/logging/SuccessfulActionCloudwatchLogging.scala
susiecoleman: fixed

app/com/gu/identity/frontend/logging/MetricsActor.scala
```diff
@@ -0,0 +1,45 @@
...
+  }
+}
+
+private sealed trait Message {}
+
+private object SignIn extends Message
+private object Register extends Message
+
+private object MetricsLoggingActor {
+  private val system = ActorSystem()
```
jamespamplin: One final thing, we should shutdown the actor system when the application shuts down. Usually the JVM will be killed when the app dies, which will kill the actor system, but when running in dev mode, the JVM is reused when the app reloads, creating a new actor system, and actor each time.

To do this, you'll need to add a `stopHook` in `ApplicationComponents` using `applicationLifecycle.addStopHook`, then shutdown the `MetricsLoggingActor` actor system from the hook.

app/com/gu/identity/frontend/configuration/Configuration.scala
susiecoleman: removed

app/com/gu/identity/frontend/logging/MetricsActor.scala
susiecoleman: added

jamespamplin:
:+1: great work!

PR #92, opened 2016-02-22
Fix: P7282.F05 User-controlled client-side redirect

markjamesbutler:
From the PEN test report:

A URL variable passed between the client and server was subsequently used to create a hyperlink and embed it in the page: https://profile.theguardian.com/register/confirm?returnUrl=http%3A%2F%2Fwww.corsaire.com By sending crafted links to users of the system, it would be possible to include third party content in the context of the application, and engage in phishing attacks, for example.


jamespamplin:
:+1: looks good to me

PR #93, opened 2016-02-23
Membership contextual sign in / registration

jamespamplin:
Support membership contextual sign in / registration flows using the URL parameter `?clientId=members`

- Membership contextual layout implemented as a "skin" - css overrides from `public/components/membership/_membership.css`
- New designs for both normal layout and membership "skin"
- Client ID propagates through entire Sign in / Registration flow
- Removed Sign in Variant B Test as results showed better conversions with A
- New MVT `RegisterMembershipStandfirst` shows upsell standfirst on membership registrations - testable using `?clientId=members&mvt_registermembershipstandfirst=a`

Also fixes:
- Minor issues with removing ReturnURL from built urls when it is the default return url (homepage)
- Catches issues with bad ReturnURLs

Screenshots:
identity-new-design-sign-in

identity-new-design-register

identity-new-design-members-sign-in

identity-new-design-members-register

identity-new-design-members-sign-in-mobile


emma-p:
:+1: Looks good to me

PR #94, opened 2016-02-24
Fix removal of http protocol from links

jamespamplin:
By default, the Google HTML Compressor will remove http protocol from href and src attributes, this change takes the defaults and modifies it so that the http protocol is preserved in the response.

Fixes issue where a Return URL requiring http will incorrectly link to https on the register confirm page due to the compressor removing the protocol from the link, eg:
https://profile.code.dev-theguardian.com/register/confirm?returnUrl=http%3A%2F%2Fm.code.dev-theguardian.com


markjamesbutler:
LGTM :+1:

PR #97, opened 2016-02-25
Live in the present

jamespamplin:
Dynamically update the year in the footer copyright.

Note: requires that the apps are restarted / deployed when the year changes.


emma-p:
:+1:

PR #98, opened 2016-02-25
Removed finished MVT tests

jamespamplin:
MVT Tests `IdentitySignInV2` and `IdentityRegisterV2` are no longer needed as we're at 100% on PROD now.


markjamesbutler:
:+1:

PR #100, opened 2016-02-26
Remove node_modules on `sbt clean`

jamespamplin:
- Removes node_modules directory on `clean` task
- Logs to the user that the node_modules have been deleted, so the next
  compile may be slow
- Log before node modules are installed for the first time, so the user
  will know it will take some time


nlindblad:
:+1:

PR #101, opened 2016-02-26
Capture legacy routes and redirect

jamespamplin:
- New Redirects controller for all redirects
- Legacy routes log request uri and referrer, redirect with permanent
  redirect code

Legacy routes for `/signin/` and `/register/` are being reported in the logs as 404. This should fix these.


nlindblad:
:+1:

PR #103, opened 2016-02-29
Make back button go back (and members logo link go to membership)

jamespamplin:
- Back button uses `returnUrl` to determine its link
- Logo link will go to membership when `clientId=members`
- `returnUrl` will fallback to membership when `clientId=members`
- Use a model for the header and footer links of the layout


nlindblad:
:+1:

AWare:
:cake:

PR #107, opened 2016-03-07
Replace send password email endpoint

emma-p:
Replaces the call to the Identity API pwd-reset endpoint (POST instead of GET)

### DIFF
--- a/app/com/gu/identity/service/client/IdentityClient.scala
+++ b/app/com/gu/identity/service/client/IdentityClient.scala
@@ -34,8 +34,14 @@ class IdentityClient extends Logging {
 
   def sendResetPasswordEmail(request: SendResetPasswordEmailApiRequest)(implicit configuration: IdentityClientConfiguration, ec: ExecutionContext): Future[Either[IdentityClientErrors, SendResetPasswordEmailResponse ]] = {
     configuration.requestHandler.handleRequest(request).map {
-      case Left(error) => Left(error)
-      case Right(r: SendResetPasswordEmailResponse) => Right(r)
+      case Left(error) => {
+        logger.info("Failed to send reset password email request")
+        Left(error)
+      }
+      case Right(r: SendResetPasswordEmailResponse) => {
+        logger.info("Successfully sent reset password email request")
+        Right(r)
+      }
       case Right(other) => Left(Seq(GatewayError("Unknown response")))
     }
   }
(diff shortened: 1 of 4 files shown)


app/com/gu/identity/service/client/IdentityClient.scala
```diff
@@ -34,8 +34,14 @@ class IdentityClient extends Logging {
...
   def sendResetPasswordEmail(request: SendResetPasswordEmailApiRequest)(implicit configuration: IdentityClientConfiguration, ec: ExecutionContext): Future[Either[IdentityClientErrors, SendResetPasswordEmailResponse ]] = {
     configuration.requestHandler.handleRequest(request).map {
-      case Left(error) => Left(error)
-      case Right(r: SendResetPasswordEmailResponse) => Right(r)
+      case Left(error) => {
+        logger.info("Failed to send reset password email request")
+        Left(error)
+      }
+      case Right(r: SendResetPasswordEmailResponse) => {
+        logger.info("Successfully sent reset password email request")
```
mchv: if we want to log more details, we probably need a bit more context like a user id, otherwise the log as it is may not be really useful?

app/com/gu/identity/service/client/IdentityClient.scala
```diff
@@ -34,8 +34,14 @@ class IdentityClient extends Logging {
 
   def sendResetPasswordEmail(request: SendResetPasswordEmailApiRequest)(implicit configuration: IdentityClientConfiguration, ec: ExecutionContext): Future[Either[IdentityClientErrors, SendResetPasswordEmailResponse ]] = {
     configuration.requestHandler.handleRequest(request).map {
-      case Left(error) => Left(error)
-      case Right(r: SendResetPasswordEmailResponse) => Right(r)
+      case Left(error) => {
+        logger.info("Failed to send reset password email request")
```
markjamesbutler: Change this to error level logging instead of info.

app/com/gu/identity/service/client/request/SendResetPasswordEmailApiRequest.scala
```diff
@@ -6,8 +6,19 @@ import com.gu.identity.service.client._
...
+  override val method = POST
+  override val headers = Iterable(
+    ApiRequest.apiKeyHeader,
+    ApiRequest.ipHeader(clientIp),
+    "Content-Type" -> "application/json")
   override val url = ApiRequest.apiEndpoint("pwd-reset/send-password-reset-email")
-  override val parameters = Seq("type" -> "reset", "email-address" -> data.email)
+  override val body = Some(SendResetPasswordEmailRequestBody(
+    `email-address` = data.email,
+    `type` = "reset"
```
markjamesbutler: Remove as discussed.

app/com/gu/identity/service/client/IdentityClient.scala
emma-p: This was more to monitor the release of the new page and get some metrics, not really to track specific journeys

mchv: ok

nlindblad:
Looks like it needs to merge in `master`. Otherwise, :+1:

PR #108, opened 2016-03-08
Add contentType config to CSRF

emma-p:
See https://www.playframework.com/security/vulnerability/20160304-CsrfBypass


adamnfish:
Looks good to me, good spot. Make sure it doesn't break any existing functionality though!

:+1:

PR #109, opened 2016-03-14
Terms and conditions

nlindblad:
This change adds the logic for showing additional terms and conditions for sign-in/registration with e.g. Teachers' Network and Guardian Jobs. It is based on the current implementation in the existing front-end project and all the URL:s are compatible with what is currently being used.

Some of the `ReturnURL` and `UrlBuilder` logic has grown a bit out of hand, so we intend to refactor those bits in a separate PR.

screen shot 2016-03-14 at 15 54 46

screen shot 2016-03-14 at 15 56 01


app/com/gu/identity/frontend/authentication/AuthenticationService.scala
```diff
@@ -0,0 +1,42 @@
...
+import play.api.mvc.{Result, Cookie, DiscardingCookie, RequestHeader}
+import play.api.mvc.Results._
+
+
+
+object AuthenticationService {
+
+  implicit def cookieNameToString(cookieName: Name): String = cookieName.toString
+
+  def authenticatedUserFor[A](request: RequestHeader, cookieDecoder: String => Option[User]): Option[AuthenticatedUser] = for {
```
jamespamplin: Type param `[A]` doesn't appear to be used here

app/com/gu/identity/frontend/authentication/AuthenticationService.scala
```diff
@@ -0,0 +1,42 @@
...
+    val cookiesToDiscard: Seq[DiscardingCookie] = Seq(
+      GuardianCookie(CookieName.gu_user_features_expiry, secure = false),
+      GuardianCookie(CookieName.gu_paying_member, secure = false),
+      GuardianCookie(CookieName.GU_U, secure = false),
+      GuardianCookie(CookieName.SC_GU_U, secure = true),
+      GuardianCookie(CookieName.GU_ID_CSRF, secure = true)
+    ).map(cookie => DiscardingCookie(cookie.name, "/", Some(cookieDomain), secure = cookie.secure))
+
+    Found(verifiedReturnUrl)
+      .withHeaders("Cache-Control" -> "no-cache", "Pragma" -> "no-cache")
```
jamespamplin: We have a [`NoCache`](https://github.com/guardian/identity-frontend/blob/master/app/com/gu/identity/frontend/controllers/Cached.scala#L46-L49) wrapper we tend to use for this

app/com/gu/identity/frontend/authentication/UserAuthenticatedAction.scala
```diff
@@ -0,0 +1,60 @@
...
+
+
+
+import scala.concurrent.Future
+import scala.util.{Success, Try}
+
+class UserAuthenticatedRequest[A](val scGuUCookie: Cookie, request: Request[A]) extends WrappedRequest[A](request)
+
+object UserAuthenticatedActionBuilder extends Logging{
+  def UserAuthenticatedAction(identityCookieDecoder: IdentityCookieDecoder) = new ActionRefiner[Request, UserAuthenticatedRequest] with ActionBuilder[UserAuthenticatedRequest] {
```
jamespamplin: We should extract `UserAuthenticatedAction` out as a class which extends `ActionRefiner` with `ActionBuilder` as will avoid creating a anonymous class

app/com/gu/identity/frontend/configuration/Configuration.scala
```diff
@@ -88,6 +92,8 @@ object Configuration {
 
     recaptchaEnabled = true,
 
+    identityCookiePublicKey = "MIIDOjCCAi0GByqGSM44BAEwggIgAoIBAQDLHjbLapKix+5Qlo3cZ6qlSXuQIG4sZxfSkD70UtdBYgXcdZa2jaNz4G/bOn+Mn+KN1KTlIupDr9+et9xy902G1JRxqAzzKgelLhEh5nTNYbOVjdi8Odg4yLqwpljTf0javSCxUE1wuOsWOZ4BNZwBpxcXeUd0MBshhLXtZiILHHLAq/4mkYBH9V/dHQXUDf7+X+kXNtR4KglvoH3zYVQ1thxUQ60rcSpkgtybHnibd5x4dVtFok4IMV05N0qW1tAhZIFWX8Dx2xene6QmEMPKsc2gJBsxWlGkX7YJiH0IMkK1D1EyzQ48TNP7oPBHeHeKtmWhFGYxsGC1XCxO3g2rAhUA8qS04q28iMQrTqgfxIZURoj5gd0CggEAUbBAySBWAY+mldTquKYHzzXylp8ocw+mpOaA+cyDvd4mucQUw/3REsZw6QlbQcGzhck2WXvJSzStAj/6PglJLMtvE/xoAk24w95iwRdaN+fYiqL0uZUc1aDXKh6BbYV1u9nlHrlv5ZfMC/dWVJCOOwYjcTfWDyTxquFDp0hT6V7sMer5Ky/xPnbAXmB/1dDomeRhKOGzf9S+aLQZR5ePV6NPLF/eh+Zb/YuHrBHI4evC2sGm6gnllyAmFdRtlT6D4/f7N7wiyBQCSpH4/XF9vGSWfRU5Wp2FgHufQd8HOXDtVnGm7KUR8A6OwHllM8zSJhI1dJtv0oD77Px0Ehea8gOCAQUAAoIBABcDIX6rJFhVPLgGrBlkZcTHqj/5Og1gRSFYSN0B7Ox3M0q3NpCaMUvrbuY8hy77M+ppHEwM8DvQ6lmMCYxUNiqTi6VGgCimZJckgWISFHXs5wqLTL5xAd8Ebohils8kljwIxV/cf+yMz9jVI+zZ8MWG5aIxDy9ThYsPIOqxpIm9tNrLyHq2LUZhM80HGSZHM2bpImg46E3OnfjV/f8mbxIEjLSaCi4kqLfSm1fiF5QzU9tnkmtxCntKvAgirOD5uWUoXsAwDktg2G5lJu9Bhp/hbzvkklp+HUBeJF+5VqnnpccYfZZT7nJ7KNNYPybGFYW6NxgTX7zptGH1chQPGvg=",
```
jamespamplin: This looks like a real key - since this is just `testConfiguration`, can we possibly just use a stubbed key instead? Or do the tests rely on a real key?

app/com/gu/identity/frontend/authentication/AuthenticationService.scala
susiecoleman: Swapped to using wrapper

app/com/gu/identity/frontend/controllers/ThirdPartyTsAndCs.scala
```diff
@@ -0,0 +1,114 @@
...
+
+import scala.concurrent.{Future, ExecutionContext}
+
+case class AddUserToGroupRequest(groupCode: String, returnUrl: Option[String])
+
+class ThirdPartyTsAndCs(identityService: IdentityService, identityCookieDecoder: IdentityCookieDecoder, config: Configuration, val messagesApi: MessagesApi, httpErrorHandler: HttpErrorHandler) extends Controller with Logging with I18nSupport {
+
+  implicit lazy val executionContext: ExecutionContext = play.api.libs.concurrent.Execution.Implicits.defaultContext
+
+  def confirmAction[A](group: String, returnUrl: Option[String], clientId: Option[String], skipThirdPartyLandingPage: Option[Boolean]) = UserAuthenticatedAction(identityCookieDecoder).async { request => {
```
jamespamplin: Is type param `[A]` required here?

app/com/gu/identity/service/client/ApiRequests.scala
```diff
@@ -25,9 +26,18 @@ object ApiRequest {
   def commonApiHeaders(trackingData: TrackingData)(implicit configuration: IdentityClientConfiguration): Iterable[(String, String)] =
     Iterable(Some(apiKeyHeader), xForwardedForIpHeader(trackingData)).flatten
 
-  def apiEndpoint(path: String)(implicit configuration: IdentityClientConfiguration) =
+  def apiSecureCookieUserHeader(cookie: Cookie) = {
```
jamespamplin: We should avoid having Play specific logic (`play.api.mvc.Cookie`) within `com.gu.identity.service.client._` classes, as plan was to potentially separate the client into a framework agnostic library.

app/com/gu/identity/frontend/authentication/UserAuthenticatedAction.scala
```diff
@@ -0,0 +1,60 @@
...
+import com.gu.identity.frontend.controllers._
+import play.api.mvc._
+import play.api.mvc.Results.SeeOther
+
+
+
+import scala.concurrent.Future
+import scala.util.{Success, Try}
+
+class UserAuthenticatedRequest[A](val scGuUCookie: Cookie, request: Request[A]) extends WrappedRequest[A](request)
```
jamespamplin: I think we should decouple this logic from the name of the Cookie, and from using the `play.api.mvc.Cookie` object. We could store a "`AuthenticatedUserCredentials`" object, or something similar, enabling flexibility for using different, or more Cookies (or other credentials) in the future. So something like:

```scala
case class UserAuthenticatedRequest[A](userCredentials: AuthenticatedUserCredentials, request: Request[A]) extends WrappedRequest[A](request)

case class AuthenticatedUserCredentials(credentials: Seq[AccessCredentials])
// could have an apply method extracting cookies from the Request object

trait AccessCredentials

case class CookieAccessCredentials(key: IdentityCookie, value: String) extends AccessCredentials

trait IdentityCookie {
  val key: String
  val value: String
}

case class SecureIdentityCookie(value: String, final val key: String = "SC_GU_U") extends IdentityCookie
```

We could also add a transformer from `SecureIdentityCookie` to a Identity API header (`X-GU-ID-FOWARDED-SC-GU-U`), to simplify creation of extraction of these headers for the API.

jamespamplin: Spoke to @nlindblad - we can make these changes separately as tech debt

app/com/gu/identity/frontend/authentication/UserAuthenticatedAction.scala
```diff
@@ -0,0 +1,60 @@
...
+            }
+          }
+        }
+        case _ => Left(SeeOther(routes.Application.signIn(Seq.empty, returnUrl, skipConfirmation, clientId.map(_.id), groupCode.map(_.getCodeValue)).url))
+      }
+    }
+  }
+  def getSC_GU_UCookie(cookies: Cookies): Option[Cookie] = cookies.get(CookieName.SC_GU_U.toString)
+
+  def getGroupCode(url: String): Option[GroupCode] = {
```
jamespamplin: These `GroupCode` extractors could go in the `GroupCode` object - can be covered separately as tech debt.

app/com/gu/identity/frontend/views/models/OAuthViewModel.scala
```diff
@@ -20,19 +20,27 @@ object OAuthProviderViewModel {
...
+      finalReturnUrl: ReturnUrl,
       skipConfirmation: Option[Boolean],
-      clientId: Option[ClientID])
+      clientId: Option[ClientID],
+      groupCode: Option[GroupCode])
       (implicit messages: Messages): OAuthProviderViewModel = {
 
     val baseUrl = configuration.identityFederationApiHost + provider.authPath
 
+    val returnUrl = groupCode match{
```
jamespamplin: I wonder if it might be easier to always return to a "identity-frontend" endpoint after OAUTH, then validate user creds / group at that point. Could help us ensure the oauth process has actually completed successfully. - Lets discuss separately as tech debt

app/com/gu/identity/frontend/authentication/AuthenticationService.scala
```diff
@@ -0,0 +1,48 @@
...
+      verifiedReturnUrl: String,
+      cookieDomain: String,
+      newCookies: Seq[Cookie] = Seq.empty): Result = {
+
+    val cookiesToDiscard: Seq[DiscardingCookie] = knownCookies.map { cookie =>
+      DiscardingCookie(name = cookie.name, path = "/", domain = Some(cookieDomain), secure = cookie.secure)
+    }
+
+    NoCache(SeeOther(verifiedReturnUrl)
+      .withHeaders("Cache-Control" -> "no-cache", "Pragma" -> "no-cache")
```
jamespamplin: You won't need this line with the `NoCache` wrapper

jamespamplin:
:+1: great work @susiecoleman and @nlindblad

nlindblad:
Thanks @jamespamplin !

PR #110, opened 2016-03-14
Improved error handling

jamespamplin:
Several goals for this PR in terms of problems to solve:
- Move explicit error handling out of logical blocks so that blocks can focus on Business logic
- Handle errors in a consistent manner so that they may be handled globally
- Ensure errors are properly logged before SignIn and Register actions redirect back to the SignIn and Register pages

Main things this does:
- Model individual errors as types extending `AppException` with `ErrorID`s used for redirects
- Explicit client errors such as `ClientRegistrationUsernameConflictError`
- New `ServiceAction` accepts a block returning `Either[ServiceExceptions, Result]` automatically handling errors as `AppException`s
- New `RedirectOnError` action will redirect to a specified route when an error occurs, automatically includes `CoreSessionParameters` (`returnUrl`, `skipConfirmation`, `clientId`, `group` code)
- Parse request bodies into `SignInActionRequestBody` and `RegisterActionRequestBody` objects for better error handling outside of the controller
- Log all errors before redirects occur with `LogOnErrorAction` - note: current master does not do this
- `UrlBuilder` now also builds out URLs with errors

Extra things:
- removed recaptcha logic from Sign in
- `ComposableActionBuilder` fixes issues with Play's `andThen` method when using `composeAction` and `composeParser` - note: need to get this into the Play Framework
- `ErrorRecoveryActionBuilder` provides a way to capture action errors using a single partial function, used by `LogOnErrorAction` and `RedirectOnError`
- Fixed issues with Play's csrf logic where it cannot read from request bodies that are not `AnyContent`. Now can also read from objects with a `CSRFTokenRequestParameter`

Note: don't review this yet until #109 has been merged


app/com/gu/identity/frontend/errors/AppException.scala
```diff
@@ -0,0 +1,63 @@
...
+
+import com.gu.identity.service.client.IdentityClientError
+
+import scala.util.control.NoStackTrace
+
+
+/**
+ * Application error.
+ */
+trait AppException extends Throwable {
```
mchv: extends [NoStackTrace](http://www.scala-lang.org/api/current/#scala.util.control.NoStackTrace) rather than `Throwable` to avoid print stack trace which is CPU costly on JVM.

mchv: Just seen that you have done it later :+1:

build.sbt
```diff
@@ -27,6 +27,8 @@ libraryDependencies ++= Seq(
 // Set logs options and default local resource for running locally (run and test)
 javaOptions ++= Seq("-Dlogs.home=logs", "-Dconfig.resource=DEV.conf")
 
+testOptions in Test += Tests.Argument("-oDF")
```
mchv: what does `oDf` settings mean?

app/com/gu/identity/frontend/errors/AppException.scala
jamespamplin: Yeah some exceptions, like "Unexpected exceptions" may benefit from a stack trace in logs, but known ones, like Invalid Credentials are known, and don't need stack traces.

build.sbt
jamespamplin: `D` adds duration to tests, and `F` reports the full stack trace when a test fails.

mchv: @ thanks!

app/com/gu/identity/frontend/services/package.scala
```diff
@@ -1,19 +1,25 @@
...
+//    val message: String
+//    val description: Option[String]
+//  }
+
+//  case class ServiceGatewayError(message: String, description: Option[String] = None) extends ServiceError {
+//    override val id: String = "error-gateway"
+//  }
+//  case class ServiceBadRequest(message: String, description: Option[String] = None) extends ServiceError {
+//    override val id: String = "error-bad-request"
+//  }
```
emma-p: Commented out code

emma-p:
As we discussed maybe add more comments, otherwise it looks good :+1:

PR #114, opened 2016-03-16
Make border-radius 18px on fields / buttons

jamespamplin:
... as per UXD feedback

signin-18px-border-radius


susiecoleman:
:+1:

PR #115, opened 2016-03-16
Fix fieldset size in Safari

jamespamplin:
Fixes the following issue where the fieldset for the signin and reg forms has an unwanted collapsed margin in Safari.

signin-fieldset-bug-safari


emma-p:
:+1:

PR #118, opened 2016-03-18
Sentry integration in JS

jamespamplin:
Initial Sentry integration in Javascript. Will log JS errors with Sentry, associating the Git hash as the Release Version for tracking errors between releases.

For example, this: [reported error](https://app.getsentry.com/the-guardian/identity-frontend-dev/issues/117568067/)


jamespamplin:
Updated config for DEV, CODE, and PROD to include Sentry DSN. Local devs will need to update their local config

susiecoleman:
:+1: Looks good to me

PR #119, opened 2016-03-18
Add CSRF user agent

markjamesbutler:
- Add user-agent logging / error message for when CSRF cookie is not present


jamespamplin:
This will conflict with the error handling work: https://github.com/guardian/identity-frontend/blob/api-error-handling/app/com/gu/identity/frontend/csrf/CSRFActions.scala#L44-L47

PR #121, opened 2016-03-18
Use nano time to measure time elapsed

mchv:
`System.nanoTime` should be used to measure time elapsed [see javadoc](https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#nanoTime).


jamespamplin:
:+1: great idea

PR #122, opened 2016-03-21
Fix handlebars for terms and conditions

nlindblad:
This change fixes an issue with the group code not being rendered in the Handlebars templates.

`GroupCode` has been changed to follow the same pattern as for `ClientId`.


jamespamplin:
:+1: good spot!

PR #123, opened 2016-03-21
Cleanup Omniture Adapter

jamespamplin:
Removes previously disabled code and unreachable code as determined as unnecessary for Omniture tracking.


emma-p:
:+1:

PR #125, opened 2016-03-23
Add inactivity alarms

susiecoleman:
Add alarms based on the cloudwatch metrics for successful signins and registrations. Alarms should trigger if we see an unexpected lack of signins or registrations. 

- Signin alarm triggers after 20 minutes without any sign ins
- Register alarm triggers after an hour without any registrations

These are based on 4 times the longest period of inactivity seen in the metrics data over the past 2 weeks. These numbers may need altering as we gather more data.


nlindblad:
:+1:

PR #126, opened 2016-03-23
Handle Social and Rate-limiting errors

jamespamplin:
Handle fbError and Rate limiting errors from the identity api.


susiecoleman:
:+1: Looks good to me

PR #127, opened 2016-03-24
Fixes for a few client-side errors reported in Sentry

jamespamplin:
- Ignore all errors from Ophan (same as dotcom)
- Catch all sessionStorage errors and perform checks before using it
- Retry loading of Ophan script, in case of network error


nlindblad:
:+1:

PR #133, opened 2016-04-12
#1202 - Register - Display error messages per field

jranks123:
Move error messages to underneath the correct field on the register page
- classify errors into types depending on what field they relate to
- show them underneath the correct field on the reg page

and 

Previously if the user entered an invalid email, the generic "one or more inputs
was not valid, please try again" message was displayed.
- Ensure that if the user enters an invalid email address, they are shown an error saying ""Invalid email address; please try again."


nlindblad:
👍

PR #135, opened 2016-04-14
Cloudwatch small datapoints

emma-p:
Sends a data point with a very small value to the CloudWatch Signin metric every 10 seconds. This makes sure the SigninInactivity alarm is never into an "INSUFFICIENT_DATA" state. Also updates the alarm to include the stage dimension


nlindblad:
👍

PR #137, opened 2016-04-19
Fix broken Guardian T&C's links

nlindblad:
Oddly the vals in `TermsViewModel` did not propagate properly to HandleBars.

This makes them part of the case class, which ensures they get through correctly.


emma-p:
👍

PR #139, opened 2016-04-25
Remove permission a/b test

markjamesbutler:
Removed test and defaulted back to control as result was not conclusive.


nlindblad:
👍

PR #141, opened 2016-04-29
Smartlock for passwords

markjamesbutler:
Implementation of SmartLock functionality providing cross device sign-in.  

[Smart Lock](https://developers.google.com/identity/smartlock-passwords/android/associate-apps-and-sites#associate_your_app_with_your_website)

### DIFF
--- a/public/components/signin-form/signin-form.js
+++ b/public/components/signin-form/signin-form.js
@@ -30,7 +30,7 @@ class SignInFormModel {
     this.state.save( email );
   }
 
-  smartLockSubmit() {
+  smartLockSetupOnSubmit() {
 
     const formElement = this.formElement.elem;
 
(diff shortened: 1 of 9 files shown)


public/components/signin-form/signin-form.js
```diff
@@ -29,7 +30,52 @@ class SignInFormModel {
...
+
+  smartLock() {
+    navigator.credentials.get({
+        password: true,
+      })
+      .then(c => {
+      if (c instanceof PasswordCredential)
+    {
+      c.additionalData = new FormData(document.querySelector('#signin-form'));
+      c.idName = "email";
```
nlindblad: For completeness, it might also be worth setting the `passwordName`, even though it defaults to `password` [1]

1. https://developer.mozilla.org/en-US/docs/Web/API/PasswordCredential/passwordName

public/components/signin-form/signin-form.js
```diff
@@ -29,7 +30,52 @@ class SignInFormModel {
...
+
+    const formElement = this.formElement.elem;
+
+    if (navigator.credentials) {
+      var c = new PasswordCredential(formElement);
+      fetch(formElement.action, {credentials: c, method: 'POST'})
+        .then(r => {
+        if (r.type == 'opaqueredirect' && !r.url.endsWith('/actions/signin')) {
+          navigator.credentials.store(c).then(_ => {
+            window.location = getElementById( 'signin_returnUrl' ).value();
```
nlindblad: Would it be worth turning this into its own method, as it is called twice?

public/components/signin-form/signin-form.js
```diff
@@ -29,7 +30,52 @@ class SignInFormModel {
...
+          });
+        }
+        })
+      ;
+    }
+
+  }
+
+  smartLock() {
```
nlindblad: Does this need some feature detection? Looks like it will always be called when the `SignInFormModel` is initialised.

public/components/signin-form/signin-form.js
```diff
@@ -29,7 +30,52 @@ class SignInFormModel {
...
+      if (c instanceof PasswordCredential)
+    {
+      c.additionalData = new FormData(document.querySelector('#signin-form'));
+      c.idName = "email";
+      fetch("/actions/signin", {credentials: c, method: 'POST'})
+        .then(r => {
+        if (r.type == 'opaqueredirect')
+      {
+        navigator.credentials.store(c).then(_ => {
+          window.location = getElementById('signin_returnUrl').value();
```
nlindblad: See previous comment about moving this to a separate function.

public/components/signin-form/signin-form.js
```diff
@@ -29,7 +30,52 @@ class SignInFormModel {
...
+        password: true,
+      })
+      .then(c => {
+      if (c instanceof PasswordCredential)
+    {
+      c.additionalData = new FormData(document.querySelector('#signin-form'));
+      c.idName = "email";
+      fetch("/actions/signin", {credentials: c, method: 'POST'})
+        .then(r => {
+        if (r.type == 'opaqueredirect')
```
nlindblad: I'm guessing we're not checking `!r.url.endsWith('/actions/signin'` here because we expect the auto-submit to always succeed? Is that always the case? What if a user changes their password or e-mail?

public/components/signin-form/signin-form.js
```diff
@@ -29,7 +30,52 @@ class SignInFormModel {
     this.state.save( email );
   }
 
+  smartLockSubmit() {
```
nlindblad: Maybe name it something that hints at the fact that this is where it gets set up? `smartLockSetupOnSubmit` or something similar. `smartLock()` also does a form submit, so it gets a bit confusing.

markjamesbutler: 👍

PR #145, opened 2016-05-05
Enable Sentry JS breadcrumbs

mario-galic:
@markjamesbutler 

http://blog.getsentry.com/2016/05/04/breadcrumbs.html


markjamesbutler:
Awesome 👍

PR #146, opened 2016-05-10
Add register diagnostic context

mario-galic:
@nlindblad 

[Trello ticket](https://trello.com/c/C5jQqvVz/8-user-could-not-be-logged-in-after-registering)


nlindblad:
👍

PR #147, opened 2016-05-12
Smart lock password reset fix

markjamesbutler:
Update to smart lock sign-in mechanism to remove infinite loop on password reset.  Prior to this PR the smart lock JS would continually attempt to execute sign-in if the credientials were incorrect.  I.e. The user has changed their password.

Additionally, specific cloud watch logging has been added for smart lock sign-ins.

### DIFF
--- a/public/components/reset-password/reset-password.js
+++ b/public/components/reset-password/reset-password.js
@@ -1,0 +1,3 @@
+/**
+ * Created by mbutler on 05/05/2016.
+ */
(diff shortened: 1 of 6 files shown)


public/components/reset-password/reset-password.js
```diff
@@ -0,0 +1,3 @@
+/**
```
nlindblad: Did you mean to check this in?

nlindblad:
👍

public/components/reset-password/reset-password.js
markjamesbutler: Nah, its a mistake.  The password reset smart lock update needs to be done in the dotcom branch as that is where the user types in the password.  I've got a separate branch to implement this.

PR #154, opened 2016-08-04
username copy tweak

ScottPainterGNM:
Based on feedback from users emailing userhelp, the users frequently ask for a change of username as they didn't realise that this would be their publically displayed name on comments.

Copy agreed with Sally and Julia to seek to reduce the overall number of these currently manual requests to Userhelp by making this more clear.


nlindblad:
👍

PR #155, opened 2016-08-15
Check email demain on registration

NathanielBennett:
Implements the email domain blacklist formerly implemented on frontend on the updated frontend app: (Here's the original PR: https://github.com/guardian/frontend/pull/6730). This is at the requesr of infosec and the mod-tools guys. 

@jamespamplin @nlindblad Can you guys take a look at this?  I've kind of shoe horned something a little bit nasty here. Really, I'd like to be able to make the job and the S3 stuff singletons that I can pass the configuration into, but I can't quite find a way of doing that. Thanks

### DIFF
--- a/app/com/gu/identity/frontend/configuration/FrontendApplicationLoader.scala
+++ b/app/com/gu/identity/frontend/configuration/FrontendApplicationLoader.scala
@@ -7,7 +7,7 @@ import com.gu.identity.frontend.controllers._
 import com.gu.identity.frontend.csrf.CSRFConfig
 import com.gu.identity.frontend.errors.ErrorHandler
 import com.gu.identity.frontend.filters.{Filters, HtmlCompressorFilter, SecurityHeadersFilter}
-import com.gu.identity.frontend.jobs.BlockedEmailDomainList
+import com.gu.identity.frontend.jobs.BlockedEmailDomainListJob
 import com.gu.identity.frontend.logging.{MetricsLoggingActor, SentryLogging, SmallDataPointCloudwatchLogging}
 import com.gu.identity.frontend.services.{GoogleRecaptchaServiceHandler, IdentityService, IdentityServiceImpl, IdentityServiceRequestHandler}
 import com.gu.identity.frontend.utils.ExecutionContexts
@@ -31,12 +31,6 @@ class FrontendApplicationLoader extends ApplicationLoader with ExecutionContexts
 
   def load(context: Context) = {
     val app = new ApplicationComponents(context).application
-
-    app.actorSystem.scheduler.schedule(
-      Duration.create(0, TimeUnit.SECONDS),
-      Duration.create(30, TimeUnit.MINUTES),
-      BlockedEmailDomainList
-    )
     new HandlebarsPlugin(app)
     app
   }
@@ -83,6 +77,8 @@ class ApplicationComponents(context: Context) extends BuiltInComponentsFromConte
   if (environment.mode == Mode.Prod) {
     new SmallDataPointCloudwatchLogging(actorSystem).start
   }
+
+  new BlockedEmailDomainListJob(actorSystem, frontendConfiguration).start
 
   applicationLifecycle.addStopHook(() => terminateActor()(defaultContext))
 
--- a/app/com/gu/identity/frontend/configuration/FrontendApplicationLoader.scala
+++ b/app/com/gu/identity/frontend/configuration/FrontendApplicationLoader.scala
@@ -1,26 +1,45 @@
 package com.gu.identity.frontend.configuration
 
-import com.gu.identity.cookie.{IdentityKeys, IdentityCookieDecoder}
+import java.util.concurrent.TimeUnit
+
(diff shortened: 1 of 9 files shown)


app/com/gu/identity/frontend/configuration/FrontendApplicationLoader.scala
```diff
@@ -1,26 +1,42 @@
...
+class FrontendApplicationLoader extends ApplicationLoader with ExecutionContexts {
+
   def load(context: Context) = {
     val app = new ApplicationComponents(context).application
+
+    app.actorSystem.scheduler.schedule(
+      Duration.create(0, TimeUnit.SECONDS),
+      Duration.create(30, TimeUnit.MINUTES),
+      BlockedEmailDomainList
+    )
```
jamespamplin: Few things here, we could move the initialisation of the Job to a class called something like `BlockedEmailDomainListJob`, under something like a `def start()` method. This class should be initialised in the `ApplicationComponents` class, here we have access to the app's `Configuration` and actor system - this is similar to how the `SmallDataPointCloudwatchLogging` actor is setup. 

Updating the agent with the data from S3 can then be performed in the new `BlockedEmailDomainListJob` which has access to the Config object, where you can specify the bucket and key config, and pass that through to the S3 object as a parameter on the `get` method.

Hope this makes sense. Give me a buzz if you'd like to pair on it.

NathanielBennett: @jamespamplin . Thanks, James, that makes sense. I'll give it a go and give you a shout if I get stuck ,,

app/com/gu/identity/frontend/services/S3.scala
```diff
@@ -0,0 +1,51 @@
...
+       logger.info(s"S3 got ${result.getObjectMetadata.getContentLength} bytes from ${result.getKey}")
+
+       try {
+         Some(action(result))
+       }
+       catch {
+         case ex: Exception => throw ex
+       }
+       finally { result.close() }
+     } catch {
```
jamespamplin: Nested try / catches here, can we possibly use one? you can just do `try { ... } finally { ... }` without a catch block as well.

app/com/gu/identity/frontend/services/S3.scala
```diff
@@ -0,0 +1,51 @@
...
+       case ex: AmazonS3Exception if ex.getStatusCode == 404 => {
+         logger.warn(s"Not found at $bucket  - $key");
```
jamespamplin: Theres a few unneeded semicolons in your code.

app/com/gu/identity/frontend/request/RegisterActionRequestBody.scala
```diff
@@ -63,6 +67,23 @@ object RegisterActionRequestBody {
     import GroupCode.FormMappings.groupCode
     import ReturnUrl.FormMapping.returnUrl
 
+    private val EmailPattern = """(.*)@(.*)""".r
+
+    def registrationEmailAddress: Constraint[String] = Constraint[String]("constraint.email") {
+      email =>
+        if (email.matches(EmailPattern.toString())) {
+          val EmailPattern(name, domain) = email
+          BlockedEmailDomainList.getBlockedDomains.contains(domain) match {
```
jamespamplin: `BlockedEmailDomainList.isDomainBlocked(domain)` might read easier here.

app/com/gu/identity/frontend/services/S3.scala
NathanielBennett: @jamespamplin  Annoyingly, you can't in this case. The inner try wraps the action call back ( in this case reading the stream from the S3 Result, into a string ), If we remove the inner try, result can't be referenced in the outer try which wraps the S3 call. As in this problem here:  http://stackoverflow.com/questions/3587746/try-block-scope

NathanielBennett:
@jamespamplin apart from try/catch;/finally, hows that?

app/com/gu/identity/frontend/utils/ExecutionContexts.scala
```diff
@@ -1,9 +1,13 @@
...
 
+import akka.actor.ActorSystem
 import scala.concurrent.ExecutionContext
+import play.api.libs.concurrent.{Akka => PlayAkka}
+import play.api.Play
 
 trait ExecutionContexts {
 
   implicit lazy val executionContext: ExecutionContext = play.api.libs.concurrent.Execution.Implicits.defaultContext
+  lazy val actorSystem: ActorSystem = PlayAkka.system(Play.current)
```
NathanielBennett: My bad this can go

jamespamplin:
Thanks @NathanielBennett, looks good to me :+1: 

Before you merge, can you please check the perms of that bucket and ensure that the app can read that file. Thanks!

NathanielBennett:
@jamespamplin It's 6.00 now so I'm going to park this till I get back on Weds morning. Hope thats OK - I'll need SSH access to check that the instances can get to the bucket OK, right? Thanks N

NathanielBennett:
Superseded by: https://github.com/guardian/identity/pull/585

PR #157, opened 2016-08-31
Discussion register with phone number

NathanielBennett:
This alters to signin/registration to take a new clientId = "comments", and asks for a mobile phone number ( internationalised ). This s so we can guage whether this will encourage first time commenters to behave better. 

Can you guys take a look?
@piuccio ( even though half of it's your code ), @markjamesbutler, @jamespamplin, @nlindblad 

with-phone


piuccio:
Extra pictures:

* JS disabled
screen shot 2016-08-30 at 17 31 10

* JS loads
screen shot 2016-08-30 at 17 32 27

* click on `why?`
screen shot 2016-08-30 at 17 32 36

It should be progressively enhanched. The JS is simply keeping the two fields in sync (one of them is hidden)

public/components/lib/phone-field.js
```diff
@@ -0,0 +1,52 @@
+const PLUGIN_OPTIONS = {
+  initialCountry: 'gb',
+  preferredCountries: ['gb', 'us', 'au']
+};
+
+export function initPhoneField (form, countryCodeElement, localNumberElement) {
+
+  require( [ 'jquery', 'intl-tel', 'intl-tel-utils' ], ($) => {
```
piuccio: this is using `require` instead of `import` so that webpack can split the code in two chunks. The one containing `jquery` is only loaded if there's a phone field in the page

conf/DEV.conf
```diff
@@ -1,6 +1,6 @@
 include "application"
 
-identity.api.host=idapi.code.dev-theguardian.com
+identity.api.host=idapi.thegulocal.com
```
nlindblad: We should keep the default to CODE.

nlindblad:
LGTM 👍 , just two minor comments.

public/components/lib/phone-field.js
jamespamplin: Good stuff @piuccio!

jamespamplin:
:+1: looks good to me too, good stuff @NathanielBennett and @piuccio

public/components/lib/phone-field.js
nlindblad: 💯

PR #158, opened 2016-09-02
Add additional helper for Chrome native form hint

nlindblad:
**Before**:

screen shot 2016-09-02 at 13 35 03

**After**:

screen shot 2016-09-02 at 13 34 38

It was already checking the format, but the native error message is not very helpful without a `title` attribute.

Based on http://www.html5rocks.com/en/tutorials/forms/constraintvalidation/


gtrufitt:
Nice 👍

mchv:
💯 awesome quick little fix following user feedback!

PR #159, opened 2016-09-05
Remember the country code on phone validation

piuccio:
When the form is persisted in local storage the country code gets lost for two reasons.

1. The fields are update after storing in local storage
2. The stored field is never used to change the initial country

This PR fixes both, when you register and validation brings you back to the form, the country flag should be the same as what you previously selected

@guardian/identity @NathanielBennett

### DIFF
--- a/public/components/register-form/_register-form.hbs
+++ b/public/components/register-form/_register-form.hbs
@@ -47,11 +47,12 @@
                 <label class="register-form__label--country-code" for="register_field_countryCode">
                   {{ registerPageText.countryCode }}
                 </label>
+                <input type="hidden" id="register_field_countryIsoName" />
                 <select class="register-form__select--country-code" id="register_field_countryCode" type="tel"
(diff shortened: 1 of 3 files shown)


public/components/register-form/_register-form.hbs
```diff
@@ -47,11 +47,12 @@
...
+                <input type="hidden" id="register_field_countryIsoName" />
                 <select class="register-form__select--country-code" id="register_field_countryCode" type="tel"
                         name="countryCode" autocomplete="on" autocapitalize="off" autocorrect="off" spellcheck="false"/>
-              {{#each countryCodes.codes }}
-              <option value="{{ code }}">+ {{ code }}</option>
-              {{/ each }}
+                  {{#each countryCodes.codes }}
+                  <option value="{{ code }}">+ {{ code }}</option>
+                  {{/ each }}
                 </select>
```
NathanielBennett: Thanks for sorting out my bad indentation. :(

NathanielBennett:
:+1: LGTM

PR #160, opened 2016-09-09
Removing unresolved export to fix build

markjamesbutler:
This seems to break on my machine but not CI which is weird.


NathanielBennett:
:+1:

PR #165, opened 2016-09-29
Fix Riff-Raff warnings

markjamesbutler:
- publicReadAcl

_[16:40:55] DEPRECATED: publicReadAcl should be specified for an autoscaling deploy. Not setting this means that it defaults to true which is insecure and probably not what you want. It is not a good idea for artifacts to be publically available on the internet - it is much better to ensure this is set to false and your instances download the artifact from S3 using IAM instance credentials. If you are CERTAIN that this iswhat you want you can also get rid of this message by explicitly setting publicReadAcl to true._

- Upgrade sbt-riffraff-artifact to latest version to ensure we are not reliant on legacy formats.

CODE build without warnings:

screen shot 2016-09-29 at 19 29 52

### DIFF
--- a/deploy.json
+++ b/deploy.json
@@ -8,7 +8,8 @@
       ],
       "data": {
         "port": "8860",
-        "bucket": "gu-identity-frontend-dist"
+        "bucket": "gu-identity-frontend-dist",
+        "publicReadAcl": false
       }
     }
   },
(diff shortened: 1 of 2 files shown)


sihil:
This will certainly fix the issue that has been highlighted. Be aware that this will mean that the artifact is no longer publicly readable. This is find provided that the artifact is loaded onto the box using a mechanism that authenticates against S3 and the role has access to the bucket in question.

Provided you can successfully deploy this branch to CODE then 👍

PR #166, opened 2016-10-14
Move Run Step After Configuration Install Step in Docs

Ap0c:
# Problem

When setting up identity locally, trying to view the health check before installing the configuration file gives an error stating that you are missing `identity.api.key`.

# Solution

Moved the instructions to view the health check after the section on installing the configuration file. Hopefully prevents devs who are setting up identity for the first time from hitting this error.


markjamesbutler:
👍

PR #169, opened 2016-10-17
Use t2.small on PROD 

markjamesbutler:
Trusted Advisor is reporting low CPU usage on these instances therefore reducing size to small.  This will reduce costs (slightly).


jfsoul:
👍

PR #171, opened 2016-10-18
Automatically sign in users with stored credentials

rupertbates:
The aim of this PR is to remove a step from the membership 'become a supporter' flow for those users who have stored Guardian credentials.

Current implementation is to check if credentials exist and if they do then redirect to the sign in page where auto sign in will be carried out by existing code.

@markjamesbutler are you happy with this approach? It introduces an extra redirect but it has the benefit of being simple and keeping all the sign in code in one place. Alternatively I could refactor the sign in and register pages to extract the common functionality.


markjamesbutler:
LGTM 👍

markjamesbutler:
@rupertbates Would be good to make this work across the board, changes look good though. 👍

public/components/register-form/register-form.js
```diff
@@ -143,3 +144,23 @@ export function init() {
 
   return form;
 }
+
+function checkForCredentials() {
+  //Only do this for the membership registration page
+  if (!window.location.search.includes("returnUrl=https://membership.theguardian.com"))
```
rupertbates: Yes it would work for all sites I was just being cautious, I'll make the change

rupertbates:
@markjamesbutler I've made the changes you suggested, can you take a look please?

markjamesbutler:
@rupertbates Looks good to me. 👍

PR #175, opened 2016-10-20
Put the build commit id into the registration page so Prout can read it

rtyley:
Previously the manifest at https://profile-origin.theguardian.com/management/manifest
was the only way to get the Git commit id out of the profile site, unfortunately the certificate on that endpoint is only valid for `profile-beta.theguardian.com`, so Prout wouldn't read it, and declared all PRs overdue-on-PROD.

This fix is an alternative to https://github.com/guardian/identity-frontend/pull/173 ("Tell Prout not to worry about the cert on profile-origin.theguardian.com")

cc @markjamesbutler

### DIFF
--- a/.prout.json
+++ b/.prout.json
@@ -5,6 +5,6 @@
       "overdue": "15M",
       "disableSSLVerification": true
     },
-    "PROD": { "url": "https://profile-origin.theguardian.com/management/manifest", "overdue": "1H" }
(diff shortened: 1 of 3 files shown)


markjamesbutler:
👍

rtyley:
Looks good on CODE & PROD! 

CODE:
image
PROD:
image

...and you can see that Prout is now correctly reporting the PRs status as 'Seen-on-PROD'!

PR #177, opened 2016-10-31
Membership text changes

rupertbates:
Change title on the sign in and registration pages for people coming from the membership flow

'Create your Guardian account' becomes 'Create your Guardian account to become a member'
and  
'Sign in to the Guardian' becomes 'Sign in to the Guardian to become a member'.

More details:
https://trello.com/c/DdaKM6f6/84-update-text-on-create-account-sign-in-pages-to-allude-to-membership


markjamesbutler:
👍  Ideally we would make this more generic for future use but it should be ok for now.  i.e. If ("register.title" + client_id) else (""register.title")

PR #180, opened 2016-11-15
Use referrer if signout returnUrl isn't provided.

markjamesbutler:
This PR adds functionality to return the user to the page they were viewing before signing out when a return url is not provided on the signout request.  

It fixes this bug:

User navigates to: (https://www.theguardian.com/uk/business) and clicks Sign Out.  User is returned to: (https://www.theguardian.com/uk)  

Once this PR is deployed the user will be returned to: (https://www.theguardian.com/uk/business)


mchv:
@markjamesbutler nice use case of `referer` header 😃

PR #185, opened 2016-11-28
Bug fix for the title of the sign up form in the Flow B of membership.

svillafe:
## Problem

Currently, when a user wants to create a _Patron_ or a _Partner_ account and he or she is in the Flow B (`clientId=membersB`), the following identity screen is shown:

bugidendity

This lead to confusion of the user because "Supporter" is the name of one of the tiers.

## Solution

When the user clicks on "Become a Patron" or "Become a Partner", we show **Become a Member** in the title.

bugidentitysolution

### DIFF
--- a/app/com/gu/identity/frontend/models/text/RegisterText.scala
+++ b/app/com/gu/identity/frontend/models/text/RegisterText.scala
@@ -1,7 +1,8 @@
 package com.gu.identity.frontend.models.text
 
-import com.gu.identity.frontend.models.{ClientID, GuardianMembersAClientID, GuardianMembersBClientID}
+import com.gu.identity.frontend.models.{ClientID, GuardianMembersAClientID, GuardianMembersBClientID, ReturnUrl}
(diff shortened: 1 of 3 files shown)


app/com/gu/identity/frontend/models/text/RegisterText.scala
```diff
@@ -53,7 +53,7 @@ object RegisterText {
       standfirst = messages("register.standfirst"),
       title = clientId match {
         case Some(GuardianMembersAClientID) => messages("register.title.membership")
-        case Some(GuardianMembersBClientID) => messages("register.title.supporter")
+        case Some(GuardianMembersBClientID) => if(returnUrl.url contains "supporter") messages("register.title.supporter") else messages("register.title.member")
```
markjamesbutler: I think it would be better to model this as a separate client rather than varying on the string in the returnUrl.  This might be flaky otherwise.

PR #187, opened 2016-11-28
Hide username for members

rupertbates:
Membership have a requirement to autopopulate the username/displayname field on the registration page from the full name fields. 

This is to reduce the number of fields which the user has to complete and so hopefully improve conversion rate.

This PR also changes the registration username field to displayname. This is consistent with social sign in and doing so removes the uniqueness constraint which would make autopopulation difficult.


app/com/gu/identity/frontend/controllers/RegisterAction.scala
```diff
@@ -32,7 +32,7 @@ class RegisterAction(identityService: IdentityService, val messagesApi: Messages
     val trackingData = TrackingData(request, body.returnUrl.flatMap(_.toStringOpt))
     identityService.registerThenSignIn(body, clientIp, trackingData).map {
       case Left(errors) =>
-        logger.error(s"User ${body.username} could not register: $errors $trackingData")
+        logger.error(s"Could not register: $errors $trackingData")
```
mario-galic: Currently this log statement is very noisy: [Sentry event](https://sentry.io/the-guardian/identity-frontend-prod/issues/128162692/)

This is because it is logging `RegisterUsernameConflictAppException` ("Username in use") and `RegisterEmailConflictAppException` ("Email in use"), which are actually valid use cases.

Should we remove this log statement or even better filter out the above two errors with something like this:

```
case Left(errors) =>

        errors.filter(_.isInstanceOf[RegisterServiceGatewayAppException]).foreach { gatewayErr =>
          logger.error(s"Could not register: $gatewayErr $trackingData")
        }
```

This would log just 5xx, for example.

markjamesbutler: @mario-galic Would you mind raising a separate PR to fix this issue?

Ap0c:
It appears that the display name field doesn't like the space added between first and last names. The return url also gets lost on this error page, so the user is unfortunately dropped out of the supporter flow here.

doesnt-like-the-space

johnduffell:
fyi: @guardian/membership-and-subscriptions  I think this has broken the membership-frontend post deployment tests: https://travis-ci.org/guardian/membership-frontend/builds/180461572
```
acceptance.util.Browser$MissingPageElementException: Could not find WebElement with locator: register_field_username
[info]   at acceptance.util.Browser$class.setValue(Browser.scala:38)
[info]   at acceptance.pages.Register.setValue(Register.scala:8)
[info]   at acceptance.pages.Register$RegisterFields$.fillIn(Register.scala:29)
```
Taking a look now

PR #188, opened 2016-11-30
use openssl to generate a local cert

johnduffell:
The certs expired already on 10th Nov. Updating the instructions to explain how to generate your own cert.

@guardian/identity


markjamesbutler:
@johnduffell Thanks for doing this.  I'm also going to investigate if we can use lets encrypt but there are issues verifying the domain.  This is great as a work around for now.

PR #191, opened 2016-12-02
Allow Sign Out to redirect to register page

markjamesbutler:
Use case:

User is signed-in under the another account (shared device).  Allow sign-out and redirect to registration page.

N.B. Initially I attempted to use default parameters to archive this but scalac doesn't like this due to

`multiple overloaded alternatives of method apply define
default arguments.`

See [Why does the Scala compiler disallow overloaded methods with default arguments?](http://stackoverflow.com/questions/4652095/why-does-the-scala-compiler-disallow-overloaded-methods-with-default-arguments)

Therefore I added additional constructors to work around this.


JustinPinner:
That looks good from our point of view I think. Thanks 👍

PR #193, opened 2016-12-05
GA Smart Lock Sign-in

markjamesbutler:
This PR fires a GA custom metric when a user successfully authenticates using Smart Lock.  Additional PR's will be raised with more meta-data (ophanId etc) and log other events (regular sign-in etc) in the near future

### DIFF
--- a/public/components/analytics/ga.js
+++ b/public/components/analytics/ga.js
@@ -14,6 +14,15 @@ function record(gaUID) {
   ga('send', 'pageview');
 }
 
+export function customMetric(event) {
+  console.log("Custom event");
+
+  var metricValue = '1';
+  ga('send', 'event', 'category', 'action', {
+    'metric2': metricValue
+  });
+}
+
 function loadGA() {
   (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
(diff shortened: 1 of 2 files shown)


public/components/signin-form/signin-form.js
```diff
@@ -1,6 +1,7 @@
 /*global window, document*/
 
 import { getElementById, sessionStorage } from '../browser/browser';
+import { customMetric as customMetric } from '../analytics/ga';
```
SiAdcock: Could be simplified to:
```
import { customMetric } from '../analytics/ga';
```

public/components/analytics/ga.js
```diff
@@ -3,15 +3,46 @@
...
   loadGA();
-  ga('create', gaUID, 'auto');
-  ga('send', 'pageview');
+  ga('create', gaUID, 'auto', gaTracker);
+  ga(gaTracker + '.send', 'pageview');
+}
+
+function buildGoogleAnalyticsEvent(event) {
+  const category = 'identity';
+  const action = event.name;
```
SiAdcock: Probably being super-picky here, but both of these `const`s could be inlined into the `fieldsObject` to reduce indirection

SiAdcock:
Aside from some minor code style nits 👍

PR #197, opened 2016-12-19
Removed the second flow of membership

svillafe:
## Why?

The checkout A/B test has finished. Flow B won the test. Therefore, we need to remove **membersB** from identity and the changes that were under membersB identifier, now will be under members.

### Before

**Old identity flow for Membership**
identitya

**New identity flow for Membership**
identityb

## Related PRs

https://github.com/guardian/membership-frontend/pull/1416

### DIFF
--- a/app/com/gu/identity/frontend/views/models/LayoutViewModel.scala
+++ b/app/com/gu/identity/frontend/views/models/LayoutViewModel.scala
@@ -2,7 +2,7 @@ package com.gu.identity.frontend.views.models
 
 import buildinfo.BuildInfo
 import com.gu.identity.frontend.configuration.Configuration
-import com.gu.identity.frontend.models.{ClientID, GuardianJobsClientID, GuardianMembersAClientID, GuardianMembersBClientID, ReturnUrl}
+import com.gu.identity.frontend.models.{ClientID, GuardianJobsClientID, GuardianMembersClientID, ReturnUrl}
 import com.gu.identity.frontend.models.Text.{HeaderText, LayoutText}
 import com.gu.identity.frontend.models.text.FooterText
 import com.gu.identity.frontend.mvt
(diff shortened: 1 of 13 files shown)


app/com/gu/identity/frontend/views/models/LayoutViewModel.scala
```diff
@@ -157,8 +157,7 @@ object LayoutLinks {
 
   private def logoUrl(configuration: Configuration, clientId: Option[ClientID]) =
     clientId match {
-      case Some(GuardianMembersAClientID) => configuration.preferredMembershipUrl
-      case Some(GuardianMembersBClientID) => configuration.preferredMembershipUrl
+      case Some(GuardianMembersClientID) => configuration.preferredMembershipUrl
```
markjamesbutler: Can we rename this membersBaseUrl to be consistent with jobs?

app/com/gu/identity/frontend/views/models/RegisterViewModel.scala
```diff
@@ -59,8 +59,7 @@ object RegisterViewModel {
     val codes = countryCodes(clientId)
 
     val isMembership = clientId match {
```
markjamesbutler: This can be converted to a single expression.  clientId == GuardianMembersClientID

app/com/gu/identity/frontend/views/models/SignInViewModel.scala
```diff
@@ -56,8 +56,7 @@ object SignInViewModel {
     val resources = getResources(layout, recaptchaModel) ++ Seq(IndirectlyLoadedExternalResources(UrlBuilder(configuration.identityProfileBaseUrl,routes.SigninAction.signInWithSmartLock())))
 
     val isMembership = clientId match {
-      case Some(GuardianMembersAClientID) => true
-      case Some(GuardianMembersBClientID) => true
+      case Some(GuardianMembersClientID) => true
```
markjamesbutler: As previous comment.  This could also be moved to a helper object to reduce duplication.

app/com/gu/identity/frontend/views/models/LayoutViewModel.scala
svillafe: Conceptually, I think it is not the same kind of URL. Whereas in the **jobs** clientId we are using the base URL (jobs.theguardian.com), in the membership case we are using the /supporter URL.

svillafe:
It is working on **CODE**, going to **PROD**

svillafe:
It is working on**PROD**

PR #205, opened 2017-02-27
Add GA metrics for sign-in and registration

jfsoul:
This adds 2 new custom GA metrics for sign-in and registration.  This does not currently include smart-lock sign-in which is handled exclusively client side.  If this change works as expected then I will introduce a similar approach for identity-federation-api for federated sign-in and registration.

This server-side approach was discussed with @markjamesbutler, and we agreed that this method was more secure and a better user experience than attempting to do this client-side via an intermediary return page.

The GA events are sent via an actor (much like the existing cloud watch events), which have been slightly reworked to more easily support referencing multiple actors from the application loader.

The GA client ID is passed from the client-side so that we can use the GA library to extract it rather than relying on the cookie directly as explained [here](https://developers.google.com/analytics/devguides/collection/analyticsjs/cookies-user-id):

> You should not directly access the cookie analytics.js sets, as the cookie format might change in the future.


app/com/gu/identity/frontend/analytics/client/MeasurementProtocolRequest.scala
```diff
@@ -0,0 +1,82 @@
...
+      "t" -> "event",
+      "uip" -> ipAddress,
+      "ua" -> userAgent,
+      "de" -> "UTF-8",
+      "ul" -> userLanguage,
+      "dl" -> url,
+      "ec" -> "identity",
+      "cd3" -> "profile.theguardian.com",
+      "cd4" -> userAgent,
+      "cd5" -> url
```
jfsoul: These params follow fairly closely the one's that the existing client-side smart lock sign in sends - although it is missing screen resolution, colour support, page title and other client-side params.

NataliaLKB:
I am afraid I don't feel qualified right now to give this a proper review 😞 . For the little I do know about this change, it looks good to me.

DiegoVazquezNanini:
I can review this change tomorrow.

app/com/gu/identity/frontend/analytics/client/MeasurementProtocolRequest.scala
```diff
@@ -0,0 +1,82 @@
...
+
+trait MeasurementProtocolRequestBody[T <: GaClientIdRequestParameter] {
+  def apply(request: Request[T], config: Configuration): String = {
+    val params = commonBodyParameters(
+      request.body.gaClientId,
+      request.remoteAddress,
+      request.headers.get("User-Agent").getOrElse(""),
+      request.acceptLanguages.headOption.map(_.language).getOrElse(""),
+      request.host + request.uri,
+      config.gaUID
```
markjamesbutler: Small thing but I usually prefer to pass in the parameter itself rather than the entire config object so we can decouple the dependency on config.  It usually helps in testing too.

app/com/gu/identity/frontend/controllers/SigninAction.scala
```diff
@@ -1,23 +1,32 @@
...
 import play.api.i18n.{I18nSupport, MessagesApi}
 import play.api.libs.concurrent.Execution.Implicits.defaultContext
-import play.api.libs.iteratee.Enumerator
-import play.api.libs.json.{JsString, JsBoolean, JsObject}
 import play.api.mvc._
 
 
 /**
- * Form actions controller
+ * Form actions controller]
```
markjamesbutler: Is this a typo?

app/com/gu/identity/frontend/request/RegisterActionRequestBody.scala
```diff
@@ -50,6 +52,7 @@ object RegisterActionRequestBody {
 
   private def handleFormErrors(formError: FormError): AppException = formError match {
     case FormError("csrfToken", _, _) => ForgeryTokenAppException("Missing csrfToken on request")
+    case FormError("gaClientId", _, _) => MissingGaClientId("Missing GA client ID on request")
```
markjamesbutler: Are we 100% that we will always have a GA Client ID?  What is the user is blocking tracking or GA?  It might be worth making the code deal with these cases.

jfsoul: Good point, I can make it optional instead but then the question is what we do if it's not defined.  Client ID or User ID is mandatory so we would have to either generate a random UUID, not track those requests or use a User ID of some sort. I assume we would in theory be able to get a User ID back from the authenticate call, but it doesn't look simple.  What do you think?

jfsoul:
@markjamesbutler changes discussed should be finished, let me know what you think.

PR #211, opened 2017-05-05
Send GA client ID to federation API

jfsoul:
This PR adds a `gaClientId` querystring parameter to the oauth links for signin and registration.  This client ID will be added to the state parameter sent to social providers in the federation API so that it can be used after successful sign in to send an event to GA.

I've also refactored how the client ID is added to the hidden field that is submitted with regular sign in.  This never belonged in the generic `ga.js` and now resides in `signin-form.js` and `register-form.js`.  It's not as DRY, but then neither are the components.  Wherever possible I have stayed consistent with the existing patterns (unlike before).

### DIFF
--- a/public/components/oauth-cta/_oauth-cta.js
+++ b/public/components/oauth-cta/_oauth-cta.js
@@ -1,0 +1,31 @@
+import { fetchTracker } from '../analytics/ga'
+
+export default class OAuthCtaModel {
+  constructor( oAuthCtaAnchors ) {
+    this.oAuthCtaAnchors = oAuthCtaAnchors;
+    this.saveClientId();
+  }
+
+  saveClientId() {
+    fetchTracker((tracker) => {
+      // Save the GA client id to be passed with the form submission
+      const clientId = encodeURI(tracker.get( 'clientId' ));
+
+      for ( var elem of this.oAuthCtaAnchors ) {
+        const connector = window.location.search.length ? '&' : '?';
+        elem.setAttribute( 'href', `${elem.getAttribute( 'href' )}${connector}gaClientId=${clientId}` );
+      }
+    });
+  }
+
+  static fromDocument() {
+    const oAuthCtaAnchors = document.getElementsByClassName( 'oauth__cta--anchor' );
+
+    if ( oAuthCtaAnchors ) {
+      return new OAuthCtaModel( oAuthCtaAnchors );
+    }
+  }
+}
+
+
+export const init = OAuthCtaModel.fromDocument;
--- a/public/components/oauth-cta/_oauth-cta.js
+++ b/public/components/oauth-cta/_oauth-cta.js
@@ -11,7 +11,7 @@ export default class OAuthCtaModel {
       // Save the GA client id to be passed with the form submission
       const clientId = encodeURI(tracker.get( 'clientId' ));
 
-      for ( var elem of this.oAuthCtaAnchors ) {
+      for ( let elem of this.oAuthCtaAnchors ) {
         const connector = window.location.search.length ? '&' : '?';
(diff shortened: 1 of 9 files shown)


jfsoul:
@markjamesbutler did you get a chance to look at this?

public/components/oauth-cta/_oauth-cta.js
```diff
@@ -0,0 +1,31 @@
...
+    this.oAuthCtaAnchors = oAuthCtaAnchors;
+    this.saveClientId();
+  }
+
+  saveClientId() {
+    fetchTracker((tracker) => {
+      // Save the GA client id to be passed with the form submission
+      const clientId = encodeURI(tracker.get( 'clientId' ));
+
+      for ( var elem of this.oAuthCtaAnchors ) {
```
SiAdcock: Wow, `for...of`, very nice! Why not use `let` instead of `var`?

public/components/oauth-cta/_oauth-cta.js
```diff
@@ -0,0 +1,31 @@
...
+        const connector = window.location.search.length ? '&' : '?';
+        elem.setAttribute( 'href', `${elem.getAttribute( 'href' )}${connector}gaClientId=${clientId}` );
+      }
+    });
+  }
+
+  static fromDocument() {
```
SiAdcock: Why a static method instead of a plain function?

public/components/register-form/register-form.js
```diff
@@ -47,12 +51,15 @@ class RegisterFormFields {
 
 
 class RegisterFormModel {
-  constructor( formElement, firstNameField, lastNameField, emailField, displayNameField, optionalCountryCode, optionalCountryIsoName, optionalPhoneNumber, optionalHideUsername ) {
+  constructor( formElement, firstNameField, lastNameField, emailField, displayNameField, optionalCountryCode, optionalCountryIsoName, optionalPhoneNumber, optionalHideUsername, gaClientIdElement ) {
```
SiAdcock: I know this predates your change, but this constructor would be so much nicer if it received a single object as a parameter, rather than a whole bunch of values that need to be input in this exact order.

public/components/signin-form/signin-form.js
```diff
@@ -97,9 +108,10 @@ class SignInFormModel {
   static fromDocument() {
     const form = getElementById( 'signin_form' );
     const emailField = getElementById( 'signin_field_email' );
+    const gaClientIdField = getElementById( 'signin_ga_client_id' );
 
     if ( form && emailField ) {
```
SiAdcock: Ought we null check `gaClientIdField` at this point?

public/components/oauth-cta/_oauth-cta.js
jfsoul: Purely consistency with the other components.  Consistency was definitely the main theme of this PR 😄

public/components/oauth-cta/_oauth-cta.js
jfsoul: good call, I forgot about `let` but at least I didn't put `const` 🔤

public/components/signin-form/signin-form.js
jfsoul: I think you're right that it needs null checking to avoid a potential call to `undefined.setValue(...)` later on.  As the GA tracking isn't critical to the sign in functionality I think it makes sense to null check in `saveClientId` method itself rather than stopping the whole component js from running.  Will make that change.

public/components/register-form/register-form.js
jfsoul: 100% but I think I might leave that for another PR

PR #215, opened 2017-05-25
Remove deploy.json for provisioning

jfsoul:
Another riff-raff.yaml


provisioning/riff-raff.yaml
```diff
@@ -0,0 +1,11 @@
+stacks:
+- identity
+regions:
+- eu-west-1
+deployments:
+  identity-frontend-provisioning:
+    type: aws-s3
+    parameters:
+      bucket: gu-identity-frontend-dist
+      cacheControl: max-age=600
```
DiegoVazquezNanini: @jfsoul Do you know why we need to cacheControl?

jfsoul: No it's only a bootstrap script so I suspect it is overkill and not really needed

DiegoVazquezNanini: Then we should remove it and just use the default.

PR #217, opened 2017-06-14
Use HTTPS links to theguardian.com

markjamesbutler:
These pages are now served directly on https so we can link to the https site directly.

### DIFF
--- a/test/com/gu/identity/frontend/models/ReturnUrlSpec.scala
+++ b/test/com/gu/identity/frontend/models/ReturnUrlSpec.scala
@@ -13,7 +13,7 @@ class ReturnUrlSpec extends FlatSpec with Matchers {
 
   it should "determine valid domain" in {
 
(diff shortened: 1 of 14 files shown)


test/com/gu/identity/frontend/models/ReturnUrlSpec.scala
```diff
@@ -13,7 +13,7 @@ class ReturnUrlSpec extends FlatSpec with Matchers {
 
   it should "determine valid domain" in {
 
-    validDomain(new URI("http://www.theguardian.com")) should be(true)
+    validDomain(new URI("https://www.theguardian.com")) should be(true)
     validDomain(new URI("http://jobs.theguardian.com")) should be(true)
     validDomain(new URI("http://code.dev-theguardian.com")) should be(true)
```
jfsoul: Could jobs and code also be https now?

PR #218, opened 2017-07-17
New style for the membership/support header.

svillafe:
New design for the membership skin. We get rid of the Guardian Members logo.

| Mobile | Mobile Landscape | Phablet | Tablet | Desktop | Wide |
|--------|------------------|---------|--------|---------|------|
|  identity-mobile    | identity-mobile-landscape  |identity-phablet | identity-tablet | identity-desktop  |  identity-wide   |

During this PR a new css class was created in order the with to be compatible with the header style in support-frontend. https://github.com/guardian/support-frontend/blob/master/assets/stylesheets/gu-sass/layout.scss#L16

### DIFF
--- a/public/components/membership/_membership.css
+++ b/public/components/membership/_membership.css
@@ -10,8 +10,8 @@
   }
 
   @media (--viewport-phablet) {
-    width: 66rem;
-    max-width: 66rem;
+    width: 660px;
+    max-width: 660px;
     padding-right: 0;
   }
 
@@ -20,15 +20,15 @@
   }
 
   @media (--viewport-min-desktop) {
-    max-width: 98rem;
+    max-width: 980px;
   }
 
   @media (--viewport-min-leftCol) {
-    max-width: 114rem;
+    max-width: 1140px;
   }
 
   @media (--viewport-min-wide) {
-    max-width: 130rem;
+    max-width: 1300px;
   }
 }
 
--- a/public/components/membership/_membership.css
+++ b/public/components/membership/_membership.css
@@ -1,13 +1,58 @@
 .skin-members {
   .header__logo-link {
-    background-image: inline("guardian-members-logo.svg");
-    height: height("guardian-members-logo.svg");
-    width: width("guardian-members-logo.svg");
     margin: 1rem 0 1.4rem;
+    height: 42px;
+    width: 224px;
+
+    @media (--viewport-min-mobile-landscape) {
+      height: 49px;
+      width: 261px;
+    }
+
+    @media (--viewport-min-tablet) {
+      height: 68px;
+      width:362px;
+    }
+
+  }
+
+  .header__back-link {
+    display: none;
+  }
+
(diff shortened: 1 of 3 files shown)


superfrank:
Not sure if it's the screenshot tool you're using, but the logo seems to have no padding on the right from Phablet upwards (which isn't correct).

jfsoul:
Would it be possible to use the basic grid system (https://github.com/guardian/identity-frontend/blob/master/public/components/_grid.css) this project has to reduce the number of custom media queries?  There are a few magic numbers being added here especially for width that would be good to keep to a minimum if possible.

cc: @NataliaLKB

superfrank:
I'd second that if possible for all pages. Then we can keep the logo in the consistent place, rather than that awkward jump.

svillafe:
@superfrank  yeap it was caused by the browser simulator. Below a screenshot in a wide screen  identity-wide

NataliaLKB:
@svillafe would you be able to update the PR description with updated screenshots (as you said they weren't right anyway)

public/components/membership/_membership.css
```diff
@@ -1,13 +1,62 @@
...
+    display: none;
+  }
+
+  .header__inner {
+    @extend %membership-header-container;
   }
 
   .header {
-    background-image: inline("bg-masthead-72px.png");
+
```
NataliaLKB: Nit: probably don't need this extra line

public/components/membership/_membership.css
```diff
@@ -1,13 +1,62 @@
+%membership-header-container {
```
NataliaLKB: Can we add a comment with a link to these values in the membership codebase? It will make it more clear why these values are needed and that you can't use our columns/previous classes.

NataliaLKB: Also I am curious as to whether a set padding with on the header with 100% width on header__inner would result in less css and clearer code.

svillafe:
@superfrank, @NataliaLKB I have just updated the screenshots.

public/components/membership/_membership.css
```diff
@@ -1,13 +1,65 @@
...
+
+  @media (--viewport-min-desktop) {
+    max-width: 98rem;
+  }
+
+  @media (--viewport-min-leftCol) {
+    max-width: 114rem;
+  }
+
+  @media (--viewport-min-wide) {
```
NataliaLKB: So followed the link and found these values: https://github.com/guardian/support-frontend/blob/d8b4be47062fdfca0dcc21199f1d156639b44be4/assets/stylesheets/gu-sass/breakpoints.scss#L20. Are they the correct corresponding values? 

Given all the other values in this file are `px` too, it would be good not to deviate.

svillafe:
It is working on Code. I will deploy it on PROD

svillafe:
It is working on PROD

NataliaLKB:
🎉

PR #223, opened 2017-08-30
Update provisioning

jfsoul:
Allow use of openJDK/Ubuntu 16.04 AMIs by updating upstart -> systemd.  Also:

- Cloudformation json -> yaml
- Remove separate bootstrap script and build for simplicity


mario-galic:
I've removed the corresponding obsolete 'Identity Frontend Provisioning' TC build.

PR #229, opened 2017-10-03
Migrate to TeamCity

mario-galic:
[New TeamCity identity-frontend build](https://teamcity-aws.gutools.co.uk/viewType.html?buildTypeId=Identity_Frontend)

identity-frontend has a custom made SBT plugin for building frontend assets: [FrontendBuildPlugin.scala](https://github.com/guardian/identity-frontend/blob/355a7b2f9b314cc2dc13b0de06c67dde1dd675cf/project/FrontendBuildPlugin.scala)

This plugin should install node dependencies, however in TeamCity it was unable to see the correct node environmental variables set in previous build step and it would default to system node 0.10.x. I managed to get it working by running everything from within the same bash script `tc-build.sh`.

Another oddity is that `riffRaffUpload` was unable to automatically pick up branch name for the manifest, so I had to create the following environmental variable:

image

The build now completes within 2 minutes after initial build.

The new TC builds are visible in riffraff starting with number 2000, while old onese are 1xxx:

image


build.sbt
```diff
@@ -35,25 +37,26 @@ javaOptions ++= Seq("-Dlogs.home=logs", "-Dconfig.resource=DEV.conf")
...
+mappings in Universal ++= directory("deploy")
 
-riffRaffPackageType := (packageZipTarball in Universal).value
+riffRaffPackageType := (packageBin in Universal).value
 riffRaffPackageName := name.value
 riffRaffManifestProjectName := s"identity:${name.value}"
-riffRaffBuildIdentifier := Option(System.getenv("CIRCLE_BUILD_NUM")).getOrElse("DEV")
 riffRaffUploadArtifactBucket := Option("riffraff-artifact")
 riffRaffUploadManifestBucket := Option("riffraff-builds")
+riffRaffBuildIdentifier := Option(System.getenv("BUILD_NUMBER")).getOrElse("unknown")
```
jfsoul: As discussed, this seems necessary - but maybe put in a short comment to say it is needed because of the custom bash build script.

PR #231, opened 2017-10-12
Email validation

rupertbates:
Our email validation currently uses the standard Html 5 input validation and the Play email constraint. However both of these allow dotless domains on email addresses eg. test@gmail

These are now [prohibited by ICANN](https://www.icann.org/news/announcement-2013-08-30-en) and the likelihood is that anyone registering with one has actually just made a mistake. Dotless domains also cause problems when we try to use them to create user accounts in Salesforce.

A bit of research by @jfsoul reveals that we have 1,984 accounts with dotless domains of which none are actually valid.

This PR makes out email validation stricter to disallow dotless domains (see [EmailValidationSpec](https://github.com/guardian/identity-frontend/pull/231/files#diff-ffc2d94f9a4c37dd928a56af5bf4aab0) for a more in depth discussion of what is and isn't valid)

### DIFF
--- a/test/com/gu/identity/frontend/models/EmailValidationSpec.scala
+++ b/test/com/gu/identity/frontend/models/EmailValidationSpec.scala
@@ -1,11 +1,10 @@
 package com.gu.identity.frontend.models
 
-import com.gu.identity.frontend.request.RegisterActionRequestBody.FormMapping
 import com.gu.identity.frontend.logging.Logging
 import com.gu.identity.frontend.request.RegisterActionRequestBody.FormMapping.dotlessDomainEmail
 import org.scalatest.{FlatSpec, Matchers}
-import play.api.data._
 import play.api.data.Forms._
+import play.api.data._
 
 class EmailValidationSpec extends FlatSpec with Matchers with Logging {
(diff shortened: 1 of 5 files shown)


test/com/gu/identity/frontend/models/EmailValidationSpec.scala
```diff
@@ -0,0 +1,81 @@
...
+      "email@-domain.com", //Leading dash in front of domain is invalid
+      "email@domain..com") //Multiple dot in the domain portion is invalid
+
+    val singleForm = Form(
+      single(
+        "email" -> dotlessDomainEmail
+      )
+    )
+
+    System.out.println("\nValid emails\n----------------------------")
```
jfsoul: Do you think it helps to leave the `println`s in? Might be a bit noisy - could use `withClue` if more detail on test cases is needed.

jfsoul:
@rupertbates FYI there's no continuous deployment to PROD on this project at the moment.  I'm going to update that now - will let you deploy this when you're ready though.

rupertbates:
Thanks for the pointer @jfsoul, deployed now.

PR #238, opened 2017-11-02
Change jobs register page text

walaura:
This was requested by Deloitte, adds an extra line linking to the jobs privacy policy under. 

They also slightly tweaked the wording on the first line and I just applied that to all login pages intead of making a new string but we need the clear from legal to actually do that, @jfsoul is on top of that!

picture 1


jfsoul:
Waiting for confirmation on this, but looks good so far.  There's a chance we might have the `standalonePrivacyPolicyText` on all the Ts&Cs as well - we shall see. Thanks!

jfsoul:
Looks like this might get held up for a couple of weeks while we get approval from stakeholders.  My bad, I should have scoped this one a bit more.  Let's leave this open for now anyway.

walaura:
no worries, should be trivial to merge in the future

walaura:
closing this until we get clearer information on which copy to use

PR #239, opened 2017-11-03
Turn the favicon red in dev

walaura:
Okay so this is the pettiest thing but it was driving me bananas AND also it's an amazing opportunity to learn scala – by getting one of you to look over my terrible hack and show me how to actually solve this cleanly

screen shot 2017-11-03 at 16 07 53


app/com/gu/identity/frontend/views/models/LayoutViewModel.scala
```diff
@@ -170,7 +169,7 @@ object Favicons {
 
   import LocalResource.resolveAssetUrl
 
-  private val iconFiles = Seq(
```
jfsoul: Generally we try to avoid mutation and stick to a more functional style.  In this case that simply means no `var` without very good justification 😄

jfsoul: In order to do that I would rename the dev icon and put it in a subdirectory `/dev` with the following code change:

```private val iconBaseDir = if (Environment.stage == "DEV") "components/favicons/dev" else "components/favicons"```

File: app/com/gu/identity/frontend/views/models/LayoutViewModel.scala [hunk 2]
```diff
@@ -180,6 +179,10 @@ object Favicons {
     "152x152.png"
   )
 
+  if(play.api.Play.isDev(play.api.Play.current)){
```
jfsoul: You'll notice in my other comment that I avoided using `Play.isDev`.  There is method to my madness - you have to be a bit careful using Play's `Mode` because it doesn't always correspond to the department convention of _stage_.  Better to stick to using stage rather than `Mode`.

I also seem to remember that `Play.current` is removed in a later release - complicating the migration path.

app/com/gu/identity/frontend/views/models/LayoutViewModel.scala
```diff
@@ -171,7 +171,7 @@ object Favicons {
   import LocalResource.resolveAssetUrl
 
   private val iconFiles = Seq(
-    "32x32.ico",
+    (if (Environment.stage == "DEV") "32x32-dev.ico" else "32x32.ico"),
```
walaura: @jfsoul got it down to one line! (and an import)

app/com/gu/identity/frontend/views/models/LayoutViewModel.scala
```diff
@@ -9,7 +9,7 @@ import com.gu.identity.frontend.mvt
 import com.gu.identity.frontend.mvt.{ActiveMultiVariantTests, MultiVariantTests, MultiVariantTest}
 import play.api.i18n.Messages
 import play.api.libs.json.Json
-
+import com.gu.identity.frontend.configuration.Configuration._
```
jfsoul: Technically I think you could just import `.Environment` here as opposed to `._`, but just minor.

PR #242, opened 2017-11-15
Parse email from url and use in frontend to pre-populate inputs

Calum-Campbell:
We want to populate email address input boxes when users land /signin or /register pages with certain query params. 
#### In this PR
- Parse the email from the url in the Applications.scala
- Pass the email from the backend to the handlebars template model and use this to prepopulate the email inputs.
-  Stop the javascript getting the email value from local storage if one has been passed in the model.
#### Valid Url's handled : 
- Landing directly: `https://profile.theguardian.com/register?email=foo@bar.com`

- When redirected from another identity page e.g `https://profile.theguardian.com/public/edit?email=foo@bar.com` : `https://profile.thegulocal.com/signin?returnUrl=https%3A%2F%2Fprofile.thegulocal.com%2Fpublic%2Fedit%3Femailasd%3Dfoo%40bar.com` 

Relies on https://github.com/guardian/frontend/pull/18247

### DIFF
--- a/public/components/register-form/register-form.js
+++ b/public/components/register-form/register-form.js
@@ -27,7 +27,10 @@ class RegisterFormFields {
   setValues( { firstName, lastName, email, displayName, optionalCountryCode, optionalCountryIsoName, optionalPhoneNumber } = {} ) {
     this.firstName.setValue( firstName );
     this.lastName.setValue( lastName );
-    this.email.setValue( email );
+    // If we don't receive an email from the backend model use local storage
+    if(this.email.value().length === 0){
+      this.email.setValue( email );
+    }
     this.displayName.setValue( displayName );
     if (this.optionalPhoneNumber) {
       this.optionalPhoneNumber.setValue(optionalPhoneNumber);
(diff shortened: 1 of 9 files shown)


jfsoul:
I was assuming that frontend was going to redirect to signin or register with an email param i.e. `https://profile.thegulocal.com/signin?returnUrl=https%3A%2F%2Fprofile.thegulocal.com%2Fpublic%2Fedit%3Femailasd%3Dfoo%40bar.com&email=foo@bar.com`

I don't think we should be taking the email from within the `returnUrl`.

Calum-Campbell:
@jfsoul Currently comes via the returnUrl as didn't change the redirect function in https://github.com/guardian/frontend/pull/18224 , probably better to be passed separately as its own query param though yep fair point.

public/components/register-form/register-form.js
```diff
@@ -27,7 +27,10 @@ class RegisterFormFields {
   setValues( { firstName, lastName, email, displayName, optionalCountryCode, optionalCountryIsoName, optionalPhoneNumber } = {} ) {
     this.firstName.setValue( firstName );
     this.lastName.setValue( lastName );
-    this.email.setValue( email );
+    // If we don't receive an email from the backend model use local storage
+    if(this.email.value().length === 0){
+      this.email.setValue( email );
+    }
```
jfsoul: minor: would be nice if there was consistency between register and signin on where this logic is i.e. it lives in `setValues` here and `loadState` in signin.  The files have different structure anyway though, so maybe not worth worrying about.

PR #266, opened 2018-01-08
Use latest consents model

jfsoul:


app/com/gu/identity/frontend/configuration/Configuration.scala
```diff
@@ -85,7 +85,7 @@ object Configuration {
 
       useSnow = false,
 
-      collectSignupConsents = true,
+      collectSignupConsents = false,
```
walaura: can you set this flag to true on `testConfiguration` too? Not quite sure what that's used for but best to keep them in sync

public/components/register-form/_form-fields.hbs
```diff
@@ -78,30 +78,13 @@
 {{#if shouldCollectConsents }}
```
walaura: i thought we weren't gonna collect any consents on signup?

jfsoul: We can remove these later, but just thought I'd leave in the example for now.

app/com/gu/identity/frontend/configuration/Configuration.scala
jfsoul: done

public/components/register-form/_form-fields.hbs
```diff
@@ -78,30 +78,13 @@
...
 
   {{#if shouldCollectV2Consents}}
-    <div class="register-form__control--gnm-marketing">
+    <div class="register-form__control--supporter-consent">
```
mario-galic: These new classes seem to be missing from *.css files. Should they be added to `_register-form.css`?

jfsoul: Yes that's a good point - I have renamed them arbitrarily, but did not worry about the CSS because they will be hidden anyway.  I suggest we tackle this in another PR after launch.

PR #279, opened 2018-01-15
Make [Sign up] link more prominent

walaura:
Move the signup link to the header, from its original location down at the footer

screen shot 2018-01-15 at 13 31 27


Calum-Campbell:
@walaura I guess @stephanfowler might want to see it but makes sense to me!

stephanfowler:
+1 looks good.

While you're here, why not put "set or reset password" after "Password" (eg on the same line as it)

walaura:
@stephanfowler that is marginally harder :(

PR #280, opened 2018-01-16
Better messaging on expired token page

Calum-Campbell:
- Better messaging on expired token page, not just generic 404
- In future PR will integrate an endpoint in Identity so we can have a "resend link" button on this page.

### Old
image

### New 
image

### DIFF
--- a/app/com/gu/identity/frontend/controllers/ConsentController.scala
+++ b/app/com/gu/identity/frontend/controllers/ConsentController.scala
@@ -1,7 +1,7 @@
 package com.gu.identity.frontend.controllers
 
 import com.gu.identity.frontend.configuration.Configuration
-import com.gu.identity.frontend.errors.ErrorIDs.UnauthorizedConsentTokenGatewayErrorID
+import com.gu.identity.frontend.errors.ErrorIDs.UnauthorizedConsentTokenErrorID
 import com.gu.identity.frontend.errors.{NotFoundError, UnauthorizedError}
 import com.gu.identity.frontend.logging.Logging
 import com.gu.identity.frontend.services.IdentityService
@@ -25,13 +25,8 @@ class ConsentController(
     identityService.processConsentToken(consentToken).map {
       case Right(playCookies) =>
         Redirect("/consents/thank-you").withCookies(playCookies: _*)
-      case Left(err) =>
-        err.id match {
-          case (UnauthorizedConsentTokenGatewayErrorID) =>
-            renderErrorPage(configuration, UnauthorizedError("Invalid token."), Unauthorized.apply)
-          case _ =>
-            renderErrorPage(configuration, NotFoundError("The requested page was not found."), NotFound.apply)
-        }
+      case Left(err) if err.id == UnauthorizedConsentTokenErrorID => renderErrorPage(configuration, UnauthorizedError("Invalid token."), Unauthorized.apply)
+      case Left(_) => renderErrorPage(configuration, NotFoundError("The requested page was not found."), NotFound.apply)
     }.recover {
       case NonFatal(e) =>
         logger.error("Failed to process consent token", e)
--- a/app/com/gu/identity/frontend/controllers/ConsentController.scala
+++ b/app/com/gu/identity/frontend/controllers/ConsentController.scala
@@ -1,7 +1,7 @@
 package com.gu.identity.frontend.controllers
 
 import com.gu.identity.frontend.configuration.Configuration
-import com.gu.identity.frontend.errors.NotFoundError
+import com.gu.identity.frontend.errors.{UnauthorizedError, NotFoundError}
 import com.gu.identity.frontend.logging.Logging
 import com.gu.identity.frontend.services.IdentityService
 import com.gu.identity.frontend.views.ViewRenderer._
@@ -24,12 +24,12 @@ class ConsentController(
     identityService.processConsentToken(consentToken).map {
       case Right(playCookies) =>
         Redirect("/consents/thank-you").withCookies(playCookies: _*)
-      case Left(_) =>
-        renderErrorPage(configuration, NotFoundError("The requested page was not found."), NotFound.apply)
+      case Left(_) =>{
+        renderErrorPage(configuration, UnauthorizedError("Invalid URL token."), Unauthorized.apply)}
     }.recover {
-      case NonFatal(e) =>
+      case NonFatal(e) => {
         logger.error("Failed to process consent token", e)
-        renderErrorPage(configuration, NotFoundError("The requested page was not found."), NotFound.apply)
+        renderErrorPage(configuration, NotFoundError("The requested page was not found."), NotFound.apply)}
     }
   }
 }
--- a/app/com/gu/identity/frontend/controllers/ConsentController.scala
+++ b/app/com/gu/identity/frontend/controllers/ConsentController.scala
@@ -1,7 +1,8 @@
 package com.gu.identity.frontend.controllers
 
 import com.gu.identity.frontend.configuration.Configuration
-import com.gu.identity.frontend.errors.{UnauthorizedError, NotFoundError}
+import com.gu.identity.frontend.errors.ErrorIDs.UnauthorizedConsentTokenGatewayErrorID
+import com.gu.identity.frontend.errors.{NotFoundError, UnauthorizedError}
 import com.gu.identity.frontend.logging.Logging
 import com.gu.identity.frontend.services.IdentityService
 import com.gu.identity.frontend.views.ViewRenderer._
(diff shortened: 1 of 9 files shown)


app/com/gu/identity/frontend/controllers/ConsentController.scala
```diff
@@ -24,12 +24,12 @@ class ConsentController(
     identityService.processConsentToken(consentToken).map {
       case Right(playCookies) =>
         Redirect("/consents/thank-you").withCookies(playCookies: _*)
-      case Left(_) =>
-        renderErrorPage(configuration, NotFoundError("The requested page was not found."), NotFound.apply)
+      case Left(_) =>{
```
jfsoul: I think we'll want to explicitly look for an unauthorised response from Identity API?

app/com/gu/identity/frontend/controllers/ConsentController.scala
```diff
@@ -24,12 +24,12 @@ class ConsentController(
...
+        renderErrorPage(configuration, UnauthorizedError("Invalid URL token."), Unauthorized.apply)}
```
jfsoul: Just call it a token rather than a URL token? And don't need the braces, I prefer leaving them out in general.

conf/messages.en
```diff
@@ -123,6 +123,10 @@ errors.unexpected.title=Unexpected error: {0}
 errors.unexpected.description=Sorry - an unexpected error has occurred, please try again later.
 errors.unexpected.details=Error details: {0}
 
+errors.unauthorized.pageTitle=Invalid consent token
+errors.unauthorized.title=Expired link
+errors.unauthorized.description=Sorry the link you have used is expired, please sign up to the newsletter again and follow the email link within a day of signing up
```
jfsoul: This error could come from marketing consents, not just email newsletters, so we'll have to update this copy.

`Sorry, the link you have used has expired.  You can manage all of your email preferences after signing in or registering (here)[link].`?

app/com/gu/identity/frontend/controllers/ConsentController.scala
```diff
@@ -24,12 +25,16 @@ class ConsentController(
     identityService.processConsentToken(consentToken).map {
       case Right(playCookies) =>
         Redirect("/consents/thank-you").withCookies(playCookies: _*)
-      case Left(_) =>
-        renderErrorPage(configuration, NotFoundError("The requested page was not found."), NotFound.apply)
+      case Left(err) =>
```
jfsoul:
```
case Left(err) if err.id == UnauthorizedConsentTokenGatewayErrorID => ...
case Left(_) => ...
```
to avoid nesting pattern matches

app/com/gu/identity/frontend/errors/ErrorID.scala
```diff
@@ -102,6 +102,10 @@ object ErrorIDs {
     val key = "consent-token-error-gateway"
   }
 
+  case object UnauthorizedConsentTokenGatewayErrorID extends ErrorID {
```
jfsoul: Not sure you need `Gateway` in the name of this?

app/com/gu/identity/frontend/services/IdentityService.scala
```diff
@@ -127,7 +127,10 @@ class IdentityServiceImpl(config: Configuration, adapter: IdentityServiceRequest
   override def processConsentToken(token: String)(implicit ec: ExecutionContext): Future[Either[ServiceException, PlayCookies]] = {
     client.postConsentToken(token) map {
       case Left(errors) =>
-        Left(ConsentTokenAppException(errors.head))
+        errors.head match {
```
jfsoul: can also avoid nesting here by doing:

```
case Left(IdentityUnauthorizedError :: _) => ...
case Left(_) => ....
```

app/com/gu/identity/service/client/Errors.scala
```diff
@@ -36,10 +36,13 @@ object IdentityClientError {
     apply(statusCode, message, None, None)
 
   def apply(statusCode: Int, message: String, description: Option[String], context: Option[String] = None): IdentityClientError =
-    if (isBadRequestError(statusCode)) ClientBadRequestError(message, description, context)
+    if (isUnauthorizedError(statusCode)) IdentityUnauthorizedError
+    else if (isBadRequestError(statusCode)) ClientBadRequestError(message, description, context)
```
jfsoul: This seems like a strange way of doing this (feels like a pattern match on status code would look better) but I realise you are just following what was there.  I would change it to a pattern match still.

app/com/gu/identity/service/client/Errors.scala
```diff
@@ -69,6 +72,11 @@ case object ClientRegistrationUsernameConflictError
   val messageForReservedUser = "Reserved user name"
 }
 
+case object IdentityUnauthorizedError
+  extends AbstractIdentityClientError("Expired Token")
```
jfsoul: Will this always mean an expired token? Seems to be named more generally than that.  Think we need to update the name or the message.

conf/messages.en
```diff
@@ -123,6 +123,10 @@ errors.unexpected.title=Unexpected error: {0}
 errors.unexpected.description=Sorry - an unexpected error has occurred, please try again later.
 errors.unexpected.details=Error details: {0}
 
+errors.unauthorized.pageTitle=Invalid consent token
+errors.unauthorized.title=Expired link
+errors.unauthorized.description=Sorry, the link you have used has expired. You can manage all of your email preferences after signing in or registering <a href="https://profile.theguardian.com/email-prefs">(here)</a>.
```
jfsoul: can we pass the href in as a variable i.e. with `{0}`.  Also I think the markup `<a></a>` belongs in the template, not in the messages file.

jfsoul: (if possible)

conf/messages.en
```diff
@@ -123,6 +123,10 @@ errors.unexpected.title=Unexpected error: {0}
 errors.unexpected.description=Sorry - an unexpected error has occurred, please try again later.
 errors.unexpected.details=Error details: {0}
 
+errors.unauthorized.pageTitle=Invalid consent token
+errors.unauthorized.title=Expired link
+errors.unauthorized.description=Sorry, the link you have used has expired. You can manage all of your email preferences after signing in or registering <a href="https://profile.theguardian.com/email-prefs">here</a>.
```
jfsoul: Not sure, but github may have swallowed my comment.  Feels like the href itself should be a variable i.e. with `{0}`.  Also if possible the markup should probably be in the template not the messages file.

app/com/gu/identity/service/client/Errors.scala
Calum-Campbell: Yeah a general name is better, I was just thinking that in this case it would mean expired token but in fact this doesn't affect the final messaging or destination so is fine to be general.

conf/messages.en
Calum-Campbell: The template is general and used for other error messages without links,  I feel like this is the best solution for now so we can continue to use lean template components and not create another template or change the current one? 
I have made it a variable then passed it in as a string but is that what you meant?  I haven't made it a variable as I don't know where we would decide what to pass in.

app/com/gu/identity/frontend/models/text/ErrorPageText.scala
```diff
@@ -37,6 +38,22 @@ object BadRequestErrorPageText {
...
+  title: String,
+  description: String
+  ) extends ErrorPageText
+
+object UnauthorizedErrorPageText {
+  def apply()(implicit messages: Messages): UnauthorizedErrorPageText =
+    UnauthorizedErrorPageText(
+      pageTitle = messages("errors.unauthorized.pageTitle"),
+      title = messages("errors.unauthorized.title"),
+      description = messages("errors.unauthorized.description", "https://profile.theguardian.com/email-prefs")
```
jfsoul: Can this just be `"/email-prefs"?

conf/messages.en
jfsoul: Yeah seems sensible for now

app/com/gu/identity/service/client/Errors.scala
```diff
@@ -35,13 +35,14 @@ object IdentityClientError {
   def apply(statusCode: Int, message: String): IdentityClientError =
     apply(statusCode, message, None, None)
 
-  def apply(statusCode: Int, message: String, description: Option[String], context: Option[String] = None): IdentityClientError =
-    if (isBadRequestError(statusCode)) ClientBadRequestError(message, description, context)
-    else ClientGatewayError(message, description, context)
+  def apply(statusCode: Int, message: String, description: Option[String], context: Option[String] = None): IdentityClientError = {
```
jfsoul: I think the outer set of braces are unnecessary

PR #284, opened 2018-01-19
Update smart lock to avoid using deprecated features

jfsoul:
Some features of the credential management API (available in Chrome for smartlock) have being deprecated and we must migrate before Jan 23rd.

Notably I have decided to only support Chrome 60+. Users of older Chrome and other browsers will just not be able to use smartlock, normal sign in should work as usual though.  Will test this on CODE briefly and check in browser stack with older browsers.

Migration guide is here for reference: https://docs.google.com/document/d/154cO-0d5paDFfhN79GNdet1VeMUmELKhNv3YHvVSOh8/edit#heading=h.p0s1hn49u2f%E2%80%9D%20with%20%E2%80%9Chttps://docs.google.com/document/d/154cO-0d5paDFfhN79GNdet1VeMUmELKhNv3YHvVSOh8/edit%23

### DIFF
--- a/public/components/register-form/register-form.js
+++ b/public/components/register-form/register-form.js
@@ -183,14 +183,16 @@ export function init() {
 }
 
 function checkForCredentials() {
-  if (navigator.credentials) {
+  if (navigator.credentials && navigator.credentials.preventSilentAccess) {
     navigator.credentials.get({
-      password: true,
+      password: true
     })
       .then(c => {
         if (c instanceof PasswordCredential) {
           const link = getElementById("sign_in_page_link");
-          window.location.replace(link.elem.href);
+          if(link && link.elem.href) {
+            window.location.replace(link.elem.href);
+          }
         }
       });
   }
--- a/public/components/register-form/register-form.js
+++ b/public/components/register-form/register-form.js
@@ -185,7 +185,7 @@ export function init() {
 function checkForCredentials() {
   if (navigator.credentials && navigator.credentials.preventSilentAccess) {
     navigator.credentials.get({
-      password: true
+      password: true,
     })
       .then(c => {
         if (c instanceof PasswordCredential) {
--- a/public/components/signin-form/signin-form.js
+++ b/public/components/signin-form/signin-form.js
@@ -73,7 +73,7 @@ class SignInFormModel {
 
   smartLockSignIn(c) {
     if (this.smartLockStatus.status) {
-      let form = new FormData(document.querySelector('#signin_form'));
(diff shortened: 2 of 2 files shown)


jfsoul:
This seems to work nicely on Chrome 63+ and Chrome < 60 (just without auto sign-in)

public/components/register-form/register-form.js
```diff
@@ -183,14 +183,16 @@ export function init() {
 }
 
 function checkForCredentials() {
-  if (navigator.credentials) {
+  if (navigator.credentials && navigator.credentials.preventSilentAccess) {
     navigator.credentials.get({
-      password: true,
+      password: true
```
walaura: trailing commas are actually best practice now!

public/components/register-form/register-form.js
```diff
@@ -183,14 +183,16 @@ export function init() {
...
     })
       .then(c => {
         if (c instanceof PasswordCredential) {
           const link = getElementById("sign_in_page_link");
-          window.location.replace(link.elem.href);
+          if(link && link.elem.href) {
```
walaura: nice! is there a reason you aren't checking for `link.elem` too?

public/components/signin-form/signin-form.js
```diff
@@ -73,22 +73,27 @@ class SignInFormModel {
...
-              this.storeRedirect(c);
-              return;
-            }
-            else {
-             r.json().then(j => {
-               this.updateSmartLockStatus(false);
-               window.location = j.url;
-               return;
-             });
+      let form = new FormData(document.querySelector('#signin_form'));
```
walaura: `const`

public/components/register-form/register-form.js
jfsoul: I've never been a fan, but I am willing to go along with the crowd 😞

public/components/register-form/register-form.js
jfsoul: was fixing a specific `can't find property elem of undefined` error - and I think because of our weird DOM element wrapper `elem` is always defined if the element itself is.

public/components/signin-form/signin-form.js
jfsoul: const 👍 

Yeah, I see your point - there are a few other fields, the most important of which is the CSRF token for the request to work at all.  It's kinda arduous passing them all in, and I'm essentially just replicating what we did before (in a different place).  I'll leave as-is for now, but if you feel strongly I can fix later.

PR #285, opened 2018-01-23
Resend Consent email page

Calum-Campbell:
## Resend consent email page with resend button

- New page when an unauthorized (invalid) token is posted to Identity
- Has resend token button on the page, this send the same token but with the expiry bumped by 24 hours.
- Successfully sent a new link page as well. 

image

image

image


Calum-Campbell:
@walaura updated!

app/com/gu/identity/service/client/IdentityClient.scala
```diff
@@ -68,6 +68,20 @@ class IdentityClient extends Logging {
     }
   }
 
+  def ResendConsentToken(request: ResendConsentTokenApiRequest)(implicit configuration: IdentityClientConfiguration, ec: ExecutionContext): Future[Either[IdentityClientErrors, ResendConsentTokenResponse ]] = {
+    configuration.requestHandler.handleRequest(request).map {
+      case Left(error) =>
+        logger.error("Failed to resend consent email link")
+        Left(error)
+
+      case Right(r: ResendConsentTokenResponse) => {
```
jfsoul: Remove curly braces here for consistency.  I would also probably put the 2 error cases together i.e. put this case statement first in the pattern match.

app/com/gu/identity/service/client/IdentityClient.scala
```diff
@@ -68,6 +68,20 @@ class IdentityClient extends Logging {
...
+        logger.info("Successfully resent consent email")
```
jfsoul: Is this a helpful log message? Genuine question.

conf/routes
```diff
@@ -3,51 +3,59 @@
...
+POST       /actions/signin/smartlock                    com.gu.identity.frontend.controllers.SigninAction.signInWithSmartLock
 
-GET        /actions/signin/consents/:token  com.gu.identity.frontend.controllers.SigninAction.permissionAuth(token:String, journey: Option[String])
+GET        /actions/signin/consents/:token              com.gu.identity.frontend.controllers.SigninAction.permissionAuth(token:String, journey: Option[String])
 
-POST       /actions/register                com.gu.identity.frontend.controllers.RegisterAction.register
+POST       /actions/register                            com.gu.identity.frontend.controllers.RegisterAction.register
 
-POST       /actions/reset                   com.gu.identity.frontend.controllers.ResetPasswordAction.reset
+GET        /invalid-token/:token                        com.gu.identity.frontend.controllers.Application.invalidConsentToken(error: Seq[String] ?= Seq.empty, token: String)
```
jfsoul: All these new endpoints will need to be added to fastly for routing.

jfsoul: (unless we do some sensible prefixing like `/consents/...`)

conf/routes
```diff
@@ -3,51 +3,59 @@
...
 
-POST       /actions/csp/report              com.gu.identity.frontend.controllers.CSPViolationReporter.cspReport
+POST       /invalid-token/resend-consent                com.gu.identity.frontend.controllers.ResendConsentTokenAction.resend
 
-GET        /management/healthcheck          com.gu.identity.frontend.controllers.HealthCheck.healthCheck
+POST       /actions/consent-email-sent/resend-consent   com.gu.identity.frontend.controllers.ResendConsentTokenAction.resend
 
-GET        /.well-known/assetlinks.json     com.gu.identity.frontend.controllers.DigitalAssetLinks.links
+GET        /actions/consent-email-sent/           com.gu.identity.frontend.controllers.Application.resendConsentTokenSent(error: Seq[String] ?= Seq.empty)
```
jfsoul: let's review naming of these URLs, think we could do a bit better

app/com/gu/identity/frontend/errors/ResendConsentEmailExeption.scala
```diff
@@ -0,0 +1,45 @@
...
+
+case class ResendConsentTokenBadTokenExeption(message: String)
+  extends AbstractAppException(message)
+    with BadRequestAppException
+    with ResendConsentTokenExeption {
+
+  val id = ResendConsentEmailBadTokenRequestId
+}
+
+case class ResendConsentTokenBadAppExeption(message: String)
```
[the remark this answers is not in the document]
Calum-Campbell: Using it as a generic bad exception from the form submission. Have renamed to ResendConsentTokenBadRequestAppException , this is in line with what is done with the resetPasswordActionRequestBody

app/com/gu/identity/frontend/models/text/ErrorPageText.scala
```diff
@@ -42,16 +42,14 @@ object BadRequestErrorPageText {
 case class UnauthorizedErrorPageText private(
   pageTitle: String,
   title: String,
-  description: String
-  ) extends ErrorPageText
+  description: String) extends ErrorPageText
```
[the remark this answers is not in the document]
Calum-Campbell: Yep agreed.

app/com/gu/identity/service/client/IdentityClient.scala
Calum-Campbell: Hmm I don't see the damage in putting it in?

conf/routes
Calum-Campbell: Do you mean like using an existing prefix? Like `/accept-consent/invalid-token/:token` and `/accept-consent/invalid-token/resend-consent`

conf/routes
```diff
@@ -3,51 +3,59 @@
...
+POST       /actions/register                            com.gu.identity.frontend.controllers.RegisterAction.register
 
-POST       /actions/reset                   com.gu.identity.frontend.controllers.ResetPasswordAction.reset
+GET        /invalid-token/:token                        com.gu.identity.frontend.controllers.Application.invalidConsentToken(error: Seq[String] ?= Seq.empty, token: String)
 
-POST       /actions/csp/report              com.gu.identity.frontend.controllers.CSPViolationReporter.cspReport
+POST       /invalid-token/resend-consent                com.gu.identity.frontend.controllers.ResendConsentTokenAction.resend
 
-GET        /management/healthcheck          com.gu.identity.frontend.controllers.HealthCheck.healthCheck
+POST       /actions/consent-email-sent/resend-consent   com.gu.identity.frontend.controllers.ResendConsentTokenAction.resend
```
Calum-Campbell: Actually this can go, this was when I was putting another `resend link` button on the " consent -email-sent" page

conf/routes
Calum-Campbell: Have changed to `/accept-consent/consent-email-sent/` , the reason I use the email-sent in the url is just to have continuity with the password reset URL's, which use `reset/email-sent`

app/com/gu/identity/service/client/request/ResendConsentTokenApiRequest.scala
```diff
@@ -0,0 +1,14 @@
...
+import com.gu.identity.frontend.request.ResendConsentTokenActionRequestBody
+import com.gu.identity.service.client._
+
+case class ResendConsentTokenApiRequest(data: ResendConsentTokenActionRequestBody)
+                                           (implicit configuration: IdentityClientConfiguration) extends ApiRequest {
+  override val method = POST
+  override val headers = Iterable(
+    ApiRequest.apiKeyHeader,
+    "Content-Type" -> "application/json")
+  override val url = ApiRequest.apiEndpoint(s"consent-resend-email/${data.token}")
```
jfsoul: Sorry I missed the identity PR for this - do you mind if I make a couple of comments on that?  See what you think and if necessary it will be easier to merge changes before this PR goes out.

app/com/gu/identity/service/client/IdentityClient.scala
jfsoul: Very minor things but: log noise and extra pressure on Logstash which can be somewhat unreliable.  If it's something we think we might look for in logs then definitely keep it, but it might be the same thing as looking for 200s in the request logs for this endpoint.  Your call!

app/com/gu/identity/service/client/request/ResendConsentTokenApiRequest.scala
Calum-Campbell: Sounds good to me.

conf/routes
```diff
@@ -3,51 +3,57 @@
...
+POST       /actions/signin/smartlock                    com.gu.identity.frontend.controllers.SigninAction.signInWithSmartLock
 
-GET        /actions/signin/consents/:token  com.gu.identity.frontend.controllers.SigninAction.permissionAuth(token:String, journey: Option[String])
+GET        /actions/signin/consents/:token              com.gu.identity.frontend.controllers.SigninAction.permissionAuth(token:String, journey: Option[String])
 
-POST       /actions/register                com.gu.identity.frontend.controllers.RegisterAction.register
+POST       /actions/register                            com.gu.identity.frontend.controllers.RegisterAction.register
 
-POST       /actions/reset                   com.gu.identity.frontend.controllers.ResetPasswordAction.reset
+GET        /accept-consent/invalid-token/:token         com.gu.identity.frontend.controllers.Application.invalidConsentToken(error: Seq[String] ?= Seq.empty, token: String)
```
jfsoul: Thinking about these new URLs.  In an ideal world I think we would have:
```
GET /consent-token/:consentToken/accept
GET /consent-token/:consentToken/invalid
POST /consent-token/resend
GET /consent-token/resent
```

but I realise that requires changing elsewhere because of the existing route (first one above).

Calum-Campbell: Yeah this is better, agreed. I have duplicated the top one so that existing systems (i.e frontend widgets etc) will still work.

app/com/gu/identity/frontend/controllers/ConsentController.scala
```diff
@@ -25,7 +25,7 @@ class ConsentController(
     identityService.processConsentToken(consentToken).map {
       case Right(playCookies) =>
         Redirect("/consents/thank-you").withCookies(playCookies: _*)
-      case Left(err) if err.id == UnauthorizedConsentTokenErrorID => renderErrorPage(configuration, UnauthorizedError("Invalid token."), Unauthorized.apply)
+      case Left(err) if err.id == UnauthorizedConsentTokenErrorID => Redirect(s"/consent-token/$consentToken/invalid")
```
jfsoul: I think you could reference the route here instead of using a string

PR #291, opened 2018-01-30
Remove smart quote from list of valid characters

rupertbates:
'Smart quotes' are invalid within an email address name portion

### DIFF
--- a/app/com/gu/identity/frontend/request/RegisterActionRequestBody.scala
+++ b/app/com/gu/identity/frontend/request/RegisterActionRequestBody.scala
@@ -74,7 +74,9 @@ object RegisterActionRequestBody {
     // This regex is based on the one used by WebKit for html email validation, documented here:
     // https://html.spec.whatwg.org/#valid-e-mail-address
     // But with the additional constraint that the domain must not be dotless
-    val dotlessDomainEmailRegex: Regex = """^([a-zA-Z0-9!#$%&’'*+\/=?^_`{|}~-]|[a-zA-Z0-9!#$%&’'*+\/=?^_`{|}~-][a-zA-Z0-9.!#$%&’'*+\/=?^_`{|}~-]*[a-zA-Z0-9!#$%&’'*+\/=?^_`{|}~-])@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?\.(?:[a-zA-Z0-9\.](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$""".r
+    // RFC spec here: https://tools.ietf.org/html/rfc2822#section-3.4.1
+    val atext = """a-zA-Z0-9!#$%&'*+\/=?^_`{|}~-"""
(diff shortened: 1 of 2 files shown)


app/com/gu/identity/frontend/request/RegisterActionRequestBody.scala
```diff
@@ -74,7 +74,9 @@ object RegisterActionRequestBody {
     // This regex is based on the one used by WebKit for html email validation, documented here:
     // https://html.spec.whatwg.org/#valid-e-mail-address
     // But with the additional constraint that the domain must not be dotless
-    val dotlessDomainEmailRegex: Regex = """^([a-zA-Z0-9!#$%&’'*+\/=?^_`{|}~-]|[a-zA-Z0-9!#$%&’'*+\/=?^_`{|}~-][a-zA-Z0-9.!#$%&’'*+\/=?^_`{|}~-]*[a-zA-Z0-9!#$%&’'*+\/=?^_`{|}~-])@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?\.(?:[a-zA-Z0-9\.](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$""".r
+    // RFC spec here: https://tools.ietf.org/html/rfc2822#section-3.4.1
+    val atext = """a-zA-Z0-9!#$%&'*+\/=?^_`{|}~-"""
```
johnduffell: great idea! although this is only valid within the square brackets [...] so I'd be inclined to include them too

app/com/gu/identity/frontend/request/RegisterActionRequestBody.scala
```diff
@@ -74,7 +74,9 @@ object RegisterActionRequestBody {
...
+    val dotlessDomainEmailRegex: Regex = s"""^([$atext]|[$atext][.$atext]*[$atext])@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?\\.(?:[a-zA-Z0-9\\.](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$$""".r
```
johnduffell: why two dollar signs at the end?

rupertbates: Because of the string interpolation

johnduffell: oh, I see, triple quotes stop working for escape sequences when using s hence the \\ and $$

johnduffell:
👍 great! 🚀

app/com/gu/identity/frontend/request/RegisterActionRequestBody.scala
rupertbates: That would make it harder to use it in the case where you need to include a dot though, you'd need two vars

app/com/gu/identity/frontend/request/RegisterActionRequestBody.scala
rupertbates: Yup 👍

PR #295, opened 2018-02-06
Resend consent token a post not a GET

Calum-Campbell:
Resend consent token should be a POST endpoint not a GET


Calum-Campbell:
Closing as probably should be a GET as this call is idempotent

PR #296, opened 2018-02-07
Eesend repermission link page

Calum-Campbell:
- Adds a "resend repermission link".
- Generalises the "link has been sent" page so repermission and consents tokens can use it

image

image


app/com/gu/identity/frontend/controllers/ResendRepermissionTokenAction.scala
```diff
@@ -0,0 +1,37 @@
...
+  identityService: IdentityService,
+  csrfConfig: CSRFConfig) extends Controller with Logging {
+
+  val redirectRoute: String = routes.Application.resendRepermissionTokenSent().url
+
+  val ResendRepermissionTokenServiceAction: ServiceActionBuilder[Request] =
+    ServiceAction andThen
+      RedirectOnError(redirectRoute) andThen
+      LogOnErrorAction(logger) andThen
+      CSRFCheck(csrfConfig)
```
rahilb: Does the order of these matter? Like will we do the `ServiceAction` before running the CSRF check for instance?

app/com/gu/identity/frontend/request/ResendTokenActionRequestBody.scala
```diff
@@ -0,0 +1,36 @@
...
+import com.gu.identity.frontend.errors._
+import com.gu.identity.frontend.request.RequestParameters.CSRFTokenRequestParameter
+import play.api.data.{Form, FormError}
+
+
+case class ResendTokenActionRequestBody(
+  token: String,
+  csrfToken: String
+)
+  extends CSRFTokenRequestParameter
```
rahilb: formatting is off.

app/com/gu/identity/frontend/controllers/ResendRepermissionTokenAction.scala
Calum-Campbell: No think it's all good. This is how we have implemented it elsewhere.

app/com/gu/identity/frontend/controllers/RepermissionController.scala
```diff
@@ -25,7 +25,7 @@ class RepermissionController(
     identityService.authenticateRepermissionToken(repermissionToken).map {
       case Right(playCookies) =>
         Redirect("/consents", request.queryString).withCookies(playCookies: _*)
-      case Left(RepermissionTokenUnauthorizedException :: _) => Redirect(routes.Application.invalidRepermissioningToken(List("error-unexpected")))
+      case Left(RepermissionTokenUnauthorizedException :: _) => Redirect(routes.Application.invalidRepermissioningToken(repermissionToken = repermissionToken))
```
jfsoul: I think named parameter passing here is probably unnecessary

app/com/gu/identity/frontend/controllers/ResendRepermissionTokenAction.scala
```diff
@@ -0,0 +1,37 @@
...
+
+  val bodyParser = ResendTokenActionRequestBody.bodyParser
+
+  def resend = ResendRepermissionTokenServiceAction(bodyParser) { request =>
+
+    identityService.resendRepermissionToken(request.body).map {
+      case Left(errors) =>
+        Left(errors)
+
+      case Right(okResponse) => Right {
```
jfsoul: (I think) this could be simplified to:

```
identityService.resendRepermissionToken(request.body).right.map {
  NoCache(SeeOther(routes.Application.resendRepermissionTokenSent().url))
}
```
Always easy to spot this when you notice you're not actually changing anything in one of the case statements (`Left` in this case).

app/com/gu/identity/frontend/errors/ResendTokenException.scala
```diff
@@ -0,0 +1,45 @@
...
+
+case class ResendTokenBadTokenException(message: String)
+  extends AbstractAppException(message)
+    with BadRequestAppException
+    with ResendTokenException {
+
+  val id = ResendTokenEmailBadTokenRequestId
+}
+
+case class ResendTokenBadRequestAppException(message: String)
```
jfsoul: I'm a little confused as to the use case for this exception.  The ID is gateway error, but it's a bad request exception.

app/com/gu/identity/frontend/services/IdentityService.scala
```diff
@@ -114,11 +115,20 @@ class IdentityServiceImpl(config: Configuration, adapter: IdentityServiceRequest
     }
   }
 
-  override def resendConsentToken(resendConsentTokenData: ResendConsentTokenActionRequestBody)(implicit ec: ExecutionContext): Future[Either[ServiceExceptions, ResendConsentTokenResponse]] = {
+  override def resendConsentToken(resendConsentTokenData: ResendTokenActionRequestBody)(implicit ec: ExecutionContext): Future[Either[ServiceExceptions, ResendTokenResponse]] = {
     val apiRequest = ResendConsentTokenApiRequest(resendConsentTokenData)
     client.resendConsentToken(apiRequest).map {
       case Left(errors) =>
-        Left(errors.map(ResendConsentTokenExeption.apply))
+        Left(errors.map(ResendTokenException.apply))
```
jfsoul: this could be simplified to left.map

app/com/gu/identity/frontend/services/IdentityService.scala
```diff
@@ -114,11 +115,20 @@ class IdentityServiceImpl(config: Configuration, adapter: IdentityServiceRequest
...
+      case Right(okResponse) => Right(okResponse)
+    }
+  }
+
+  override def resendRepermissionToken(resendRepermissionTokenData: ResendTokenActionRequestBody)(implicit ec: ExecutionContext): Future[Either[ServiceExceptions, ResendTokenResponse]] = {
+    val apiRequest = ResendRepermissionTokenApiRequest(resendRepermissionTokenData)
+    client.resendRepermissionToken(apiRequest).map {
+      case Left(errors) =>
+        Left(errors.map(RepermissionTokenAppException.apply))
```
jfsoul: same as above, this could be left.map

public/invalid-repermission-token-page.hbs
```diff
@@ -0,0 +1,27 @@
...
+
+    {{else}}
+      <p class="page-standfirst__paragraph resend-link-description">{{ text.description }}</p>
+
+      <form id="resend-token-form" class="resend-consent-form" action="/repermission-token/resend" method="GET">
+        <input type="hidden" name="token" value="{{ token }}">
+        <input type="hidden" name="{{ csrfToken.fieldName }}" value="{{ csrfToken.value }}">
+
+        <div class="resend-consent-form__control--submit">
+          <button class="resend-consent-form__submit-button" id="resend-consent_submit" type="submit">Send me a new one {{> components/icon/arrow-inline }}</button>
```
jfsoul: Should this button text be in the text model for this page?

public/components/resend-consent/_resend-consent.css
```diff
@@ -37,3 +37,9 @@
 .resend-consent-form__submit-button {
   @extend %form__button--submit;
 }
+
+.page-standfirst {
+  & .resend-link-description {
```
walaura: Is there any particular reason this one standfirst should be more spaced out than the rest? Can you make them all have the same spacing?

walaura:
Regarding the last screenshot, could that error page have a generic «Oops» header? Just to not lead with an error message

app/com/gu/identity/frontend/controllers/RepermissionController.scala
Calum-Campbell: It's just because the first parameter of that method is an ```Option[Seq[String]]``` , so need to name it here.

app/com/gu/identity/frontend/errors/ResendTokenException.scala
Calum-Campbell: Yep your right, was the wrong exception ID

app/com/gu/identity/frontend/controllers/ResendRepermissionTokenAction.scala
Calum-Campbell: It's actually a future of an Either so I would have to look like : 
```
  identityService.resendRepermissionToken(request.body).map {_.right.map{ _ =>
      NoCache(SeeOther(routes.Application.resendRepermissionTokenSent().url))
    }}
```

I feel like the current state is slightly clearer?

app/com/gu/identity/frontend/controllers/RepermissionController.scala
jfsoul: Ok so there doesn't need to be any errors passed now?

app/com/gu/identity/frontend/controllers/ResendRepermissionTokenAction.scala
jfsoul: Ah yeah, down to choice of style then.  Your call!

app/com/gu/identity/frontend/controllers/RepermissionController.scala
Calum-Campbell: Actually yeah maybe I should be passing : 
```
      case Left(RepermissionTokenUnauthorizedException :: _) => Redirect(routes.Application.invalidRepermissioningToken(List(RepermissionTokenUnauthorizedException.id.key), repermissionToken))

```

Calum-Campbell: In fact, I think I should just remove passing the error at all, as we want this error case to display the 'resend link' button and all other error cases to render a generic errorPage.

PR #298, opened 2018-02-19
redirect to staywithus

Calum-Campbell:
Redirect to new /staywithus URL for repermission journey.


Calum-Campbell:
@jfsoul The only difference between the /consents and /consents/staywithus at the moment is a green banner at the top that says "Stay with us" so it would probably be marginally better to merge this sooner rather than later. But not much in it.

image

jfsoul:
Yeah, just wondering whether that is better or worse than saying nothing 😆

walaura:
fwiw i intend to add some graphics & branding to /staywithus in the short term (think tomorrow). Only branding & no messaging so we should be fine but keep it in mind when making this call

PR #304, opened 2018-03-06
Initial two step sign in work

walaura:
Adds new GET endpoints for passwordless/two-step signin, namely:

- `/signin/start` (this will replace `/signin` once it's ready)
- `/signin/:signintype` where `signintype` is compared against a set of valid types (new,existing,guest) or errors otherwise

As a full flow this is non functional at the moment as there's backend work needed and decisions about how to tie these together to be made. For the time being I just want to get the ball rolling without affecting the existing flow

(Sorry for the huge PR, you know how this repo is. Luckily It's mostly new files)


app/com/gu/identity/frontend/models/SignInType.scala
```diff
@@ -0,0 +1,28 @@
+package com.gu.identity.frontend.models
+
+sealed trait SignInType {
```
jfsoul: I find using trait and case objects for this a bit long-winded, and you could say the same for pattern matching.  Alternative implementation (not saying this is better):

```
sealed abstract class SignInType(val id: String)

object SignInType {
  private val all: Seq[SignInType] = Seq(ExistingUser, GuestUser, NewUser)
  
  def apply(signInType: String) = all.find(_.id == signInType)
}

case object ExistingUser extends SignInType("existing")
case object GuestUser extends SignInType("guest")
case object NewUser extends SignInType("new")
```

On the other hand, if you want to keep pattern matching then note that you can avoid some LOC:

```
object SignInType {
   def apply(signInType: String): Option[SignInType] = {
       case "existing" => Some(ExistingUser)
       case "guest" => Some(GuestUser)
       case "new" => Some(NewUser)
       case _ => None
   }
 }
```

app/com/gu/identity/frontend/views/ViewRenderer.scala
```diff
@@ -53,6 +54,68 @@ object ViewRenderer {
...
+      errors = errorIds.map(ErrorViewModel.apply),
+      returnUrl = returnUrl,
+      skipConfirmation = skipConfirmation,
+      clientId = clientId,
+      group = group,
+      email = email,
+      signInType = signInType
+    )
+
+    (signInType,email) match {
```
jfsoul: Personal taste but I tend to lean away from pattern matching artificial tuples in this way although sometimes they are clearer than if..else.  Make your own mind up on this case:

```
    if(signInType.isDefined)
      if (email.isDefined) renderViewModel("two-step-signin-choices-page", model)
      else NoCache(SeeOther(routes.Application.twoStepSignIn().url))
    else
      renderErrorPage(configuration, NotFoundError("The requested page was not found."), NotFound.apply)
```

app/com/gu/identity/frontend/models/SignInType.scala
walaura: I copypasted the existing `models/GroupCode` code. I'm more than happy to switch to the leaner version on top

jfsoul: Hmm that group code implementation is unusual if you ask me 🤔

walaura: Gonna agree with you there haha

PR #306, opened 2018-03-07
Update the asset build system 

walaura:
decouples npm from sbt, no longer causing it to run on each page load but instead presenting a separate `npm run watch` command. This is way faster and helps decouple otherwise unrelated tooling, letting tasks run in parallel.

While I was there i upgraded webpack and postcss and had to mess a but with the css to get it to run in upper versions. future pr's will try to reduce the number of postcss plugins to a more manageable number (so there's no multiple styles of variables for instance)

## tested in CODE?
YES


.nvmrc
```diff
@@ -1 +1 @@
-4.5.0
+9.3.0
```
sndrs: might want the lts version here? depends if this app is likely to go quiet in the future

walaura: absolutely

PR #309, opened 2018-03-07
Add templates for all 3 signin types

walaura:
Makes all three two step sign in pages actually work. The flows and messaging are still somewhat clunky for this PR but this brings it closer to a working prototype.

### Existing user (email & pwd on file)
screen shot 2018-03-07 at 6 32 40 pm

### Guest user (email on file but has no pwd. Users from newsletters and such)
screen shot 2018-03-07 at 6 32 56 pm

### New user (email not on file)
screen shot 2018-03-07 at 6 33 03 pm

### DIFF
--- a/app/com/gu/identity/frontend/controllers/Application.scala
+++ b/app/com/gu/identity/frontend/controllers/Application.scala
@@ -48,7 +48,7 @@ class Application (configuration: Configuration, val messagesApi: MessagesApi, c
     val returnUrlActual = ReturnUrl(returnUrl, req.headers.get("Referer"), configuration, clientIdActual)
     val csrfToken = CSRFToken.fromRequest(csrfConfig, req)
     val groupCode = GroupCode(group)
-    val email : Option[String] = req.getQueryString("email").orElse(req.cookies.get("GU_SIGNIN_EMAIL").map(_.value))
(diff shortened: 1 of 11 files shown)


app/com/gu/identity/frontend/controllers/Application.scala
```diff
@@ -48,7 +48,7 @@ class Application (configuration: Configuration, val messagesApi: MessagesApi, c
     val returnUrlActual = ReturnUrl(returnUrl, req.headers.get("Referer"), configuration, clientIdActual)
     val csrfToken = CSRFToken.fromRequest(csrfConfig, req)
     val groupCode = GroupCode(group)
-    val email : Option[String] = req.getQueryString("email")
+    val email : Option[String] = req.getQueryString("email").orElse(req.cookies.get("GU_SIGNIN_EMAIL").map(_.value))
```
Calum-Campbell: In the api we use emailAddress in the query string for this. Maybe good to keep continuity, maybe doesn't matter?

walaura: Maybe we can remove this paremeter altogether to be honest, i'm not sure if it's in use anywhere

PR #315, opened 2018-03-08
Add js to the two step login sequence

walaura:
Progressive enhancement! This is using a mock url for tests while I'm waiting for the endpoint but the flow is way nicer now

2018-03-08 3_25_21 pm


public/components/two-step-signin/two-step-signin.js
```diff
@@ -0,0 +1,90 @@
...
+      resolve($new);
+    });
+    requestAnimationFrame(() => {
+      $old.classList.remove("two-step-signin__slide--visible");
+      $old.classList.add(classNames.out);
+    });
+    $new.addEventListener("animationend", () => {
+      [
+        "two-step-signin__slide--in",
+        "two-step-signin__slide--out",cont
```
jfsoul: this looks weird `cont`?

app/com/gu/identity/frontend/views/models/TwoStepSignInViewModel.scala
```diff
@@ -36,7 +36,7 @@ case class TwoStepSignInViewModel private(
   //TODO: use reverse signInWithEmail route
   actions: Map[String, String] = Map(
     "signInWithEmailAndPassword" -> routes.SigninAction.signIn().url,
-    "signInWithEmail" -> "/signin/with-email"
+    "signInWithEmail" -> "/actions/signin/with-email"
```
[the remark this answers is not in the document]
walaura: not yet because it doesn't exist! as you can see from the comment on the top it's being looked at tho

public/components/two-step-signin/two-step-signin.js
walaura: whoa. see this is one of the things #316 is gonna fix

PR #316, opened 2018-03-08
Enforce js & css linting at push time

walaura:
The build change from yesterday meant code linting wasn't being enforced anymore. This fixes it.

First of all i installed typescript (😍), the goal is to move all `.js` to `.ts` so we won't be flying blind much longer, this also gives me peace of mind dismantling the old js structure, which had a lot of boilerplate serving as type check. For now it's just there, and I'll migrate the new two step signin stuff in a separate pr.

Second of all I copied a subset of tools over from `frontend`to validate js & css code using `prettier`. This also gives us essential scaffolding to eventually put there type checks (such as push-time typescript) or any other tools. (like a more comprehensive css linter)

<div align=center>
~~<br>

<i><small>(It also introduces these annoying messages blocking pushing if anything goes wrong.)</small></i>
~~
</div>

I ran the tool on all current files so the changelog is kind of on the large side. Sorry! This is a one time thing


public/components/analytics/ga.js
```diff
@@ -53,8 +53,23 @@ function buildGoogleAnalyticsEvent(event) {
 }
 
 function loadGA() {
-  (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
-      (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
-    m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
-  })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
```
jfsoul: Thanks for fixing GA prettier 😆

public/components/analytics/ophan.js
```diff
@@ -3,48 +3,46 @@
...
 
 // Use curl loader as Ophan library is AMD
-import 'curl-amd/dist/curl/curl';
+import "curl-amd/dist/curl/curl";
 
-import { getActiveTestsAndResultsForOphan } from './mvt';
+import { getActiveTestsAndResultsForOphan } from "./mvt";
 
-const ophanScript = 'https://j.ophan.co.uk/ophan.ng.js';
+const ophanScript = "https://j.ophan.co.uk/ophan.ng.js";
```
jfsoul: are double quotes trendy now?

webpack.config.js
```diff
@@ -14,21 +15,26 @@ module.exports = {
   },
   module: {
     rules: [
+      {
+        test: /\.tsx?$/,
```
jfsoul: what's the typescript for?

walaura: the future! i feel a bit unsafe writing js without types

jfsoul: I'm no expert on typed js, but department seems to be going the direction of flow instead of typescript. Something to consider.

PR #321, opened 2018-03-12
Two step email signin

Calum-Campbell:
- Redirect to /signin/new , signin/current  or signin/guest depending on the type of user returned from Identity API
- Issues a GU_SIGNIN_EMAIL cookies that lasts for 10 seconds on the redirect. This allows us to prepopulate the form fields on the next page

### DIFF
--- a/app/com/gu/identity/frontend/authentication/CookieService.scala
+++ b/app/com/gu/identity/frontend/authentication/CookieService.scala
@@ -63,7 +63,7 @@ object CookieService {
     Seq(PlayCookie(
       name = "GU_SIGNIN_EMAIL",
       value = email,
-      maxAge = Some(10),
+      maxAge = Some(120),
       path = "/",
       domain = Some(config.identityCookieDomain),
       secure = true,
--- a/app/com/gu/identity/frontend/authentication/CookieService.scala
+++ b/app/com/gu/identity/frontend/authentication/CookieService.scala
@@ -59,4 +59,17 @@ object CookieService {
     }
   }
 
+  def signInEmailCookies(email: String)(config: Configuration): Seq[PlayCookie] = {
+    Seq(PlayCookie(
+      name = "GU_SIGNIN_EMAIL",
+      value = email,
+      maxAge = Some(10),
+      path = "/",
+      domain = Some(config.identityCookieDomain),
+      secure = true,
+      httpOnly = false
+    )
+    )
+  }
+
 }
--- a/app/com/gu/identity/frontend/authentication/CookieService.scala
+++ b/app/com/gu/identity/frontend/authentication/CookieService.scala
@@ -63,7 +63,7 @@ object CookieService {
     Seq(PlayCookie(
       name = "GU_SIGNIN_EMAIL",
       value = email,
-      maxAge = Some(120),
+      maxAge = None,
       path = "/",
       domain = Some(config.identityCookieDomain),
       secure = true,
--- a/app/com/gu/identity/frontend/controllers/SigninAction.scala
+++ b/app/com/gu/identity/frontend/controllers/SigninAction.scala
@@ -13,6 +13,7 @@ import play.api.i18n.{I18nSupport, MessagesApi}
 import play.api.libs.concurrent.Execution.Implicits.defaultContext
 import play.api.mvc._
 import Configuration.Environment._
+import com.gu.identity.frontend.authentication.CookieService
 import com.gu.tip.Tip
 
 import scala.concurrent.Future
@@ -73,16 +74,15 @@ class SigninAction(
   }
 
   def signInWithEmail = SignInEmailServiceAction(bodyParser) {
-    emailSignInAction(
-      (url, cookies) => SeeOther(url).withCookies(cookies: _*),
-      _ => metricsActor.logSuccessfulEmailSignin()
-    )
+    emailSignInAction(successfulFirstStepResponse, _ => metricsActor.logSuccessfulEmailSignin())
   }
 
-  def emailSignInAction(successResponse: (String, Seq[Cookie]) => Result, metricsLogger: (Request[SignInActionRequestBody]) => Unit) = { implicit request: Request[SignInActionRequestBody] => {
+  def emailSignInAction(successResponse: (ReturnUrl, Seq[Cookie]) => Result, metricsLogger: (Request[SignInActionRequestBody]) => Unit) = { implicit request: Request[SignInActionRequestBody] => {
+//  def emailSignInAction() = { implicit request: Request[SignInActionRequestBody] => {
 
     val body = request.body
 
+    println(body)
     lazy val returnUrl = body.returnUrl.getOrElse(ReturnUrl.defaultForClient(config, body.clientId))
 
     val successfulReturnUrl = body.groupCode match {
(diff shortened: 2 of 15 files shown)


walaura:
can the cookie last 120 seconds instead? /signin and /reset have to pick it up too

app/com/gu/identity/frontend/controllers/SigninAction.scala
```diff
@@ -115,6 +151,10 @@ class SigninAction(
     SeeOther(successfulReturnUrl.url)
       .withCookies(cookies: _*)
 
+  def successfulFirstStepResponse(userType: String, successfulReturnUrl: ReturnUrl, cookies: Seq[Cookie]): Result =
+    SeeOther(s"/signin/${userType}?returnUrl=${java.net.URLEncoder.encode(successfulReturnUrl.url, "UTF8")}")
```
walaura: this will also need to pass clientId & possibly UTM tags :(

app/com/gu/identity/frontend/views/models/TwoStepSignInViewModel.scala
```diff
@@ -36,7 +36,7 @@ case class TwoStepSignInViewModel private(
   //TODO: use reverse signInWithEmail route
```
walaura: no need for this anymore!

app/com/gu/identity/frontend/authentication/CookieService.scala
```diff
@@ -59,4 +59,17 @@ object CookieService {
     }
   }
 
+  def signInEmailCookies(email: String)(config: Configuration): Seq[PlayCookie] = {
```
jfsoul: This is basically to avoid putting `e=<email or encoded email>`?

I'm still not sure this is better enough that it justifies the extra effort and cookie pollution.

app/com/gu/identity/frontend/controllers/SigninAction.scala
```diff
@@ -43,6 +46,13 @@ class SigninAction(
       LogOnErrorAction(logger) andThen
       CSRFCheck(csrfConfig)
 
+  val SignInEmailServiceAction =
```
jfsoul: I'm not convinced we need a new one of these if it is the same as the other(s).

app/com/gu/identity/frontend/controllers/SigninAction.scala
```diff
@@ -63,6 +73,32 @@ class SigninAction(
     signInAction(successfulSmartLockSignInResponse, _ => metricsActor.logSuccessfulSmartLockSignin())
   }
 
+  def signInWithEmail = SignInEmailServiceAction(bodyParser) {
+    emailSignInAction(successfulFirstStepResponse, _ => metricsActor.logSuccessfulEmailSignin())
+  }
+
+  def emailSignInAction(successResponse: (String, ReturnUrl, Seq[Cookie]) => Result, metricsLogger: (Request[SignInActionRequestBody]) => Unit) = { implicit request: Request[SignInActionRequestBody] => {
+
+    val body = request.body
```
jfsoul: this seems a bit superfluous

app/com/gu/identity/frontend/controllers/SigninAction.scala
```diff
@@ -63,6 +73,32 @@ class SigninAction(
...
+
+    lazy val returnUrl = body.returnUrl.getOrElse(ReturnUrl.defaultForClient(config, body.clientId))
+
+    val successfulReturnUrl = body.groupCode match {
+      case Some(validGroupCode) =>
```
jfsoul: Do we know this group code is valid (just wondering about the name)?

app/com/gu/identity/frontend/controllers/SigninAction.scala
jfsoul: import java.net.URLEncoder

app/com/gu/identity/frontend/logging/MetricsActor.scala
```diff
@@ -22,6 +23,10 @@ class MetricsLoggingActor(metricsActor: ActorRef) {
     metricsActor ! SmartLockSignIn
   }
 
+  def logSuccessfulEmailSignin() = {
+    metricsActor ! EmailSignIn
```
jfsoul: This needs to be implemented below in receive

app/com/gu/identity/service/client/IdentityClient.scala
```diff
@@ -116,4 +116,18 @@ class IdentityClient extends Logging {
       case Right(other) => Left(Seq(ClientGatewayError("Unknown response")))
     }
   }
+
+  def getUserType(request: UserTypeRequest)(implicit configuration: IdentityClientConfiguration, ec: ExecutionContext): Future[Either[IdentityClientErrors, UserTypeResponse]] = {
+    configuration.requestHandler.handleRequest(request).map {
+      case Left(error) =>
+        logger.info("Failed to get user type")
```
jfsoul: error instead of info?

app/com/gu/identity/frontend/models/SignInType.scala
```diff
@@ -7,6 +7,6 @@ object SignInType {
   def apply(signInType: String) = all.find(_.id == signInType)
 }
 
-case object ExistingUser extends SignInType("existing")
+case object ExistingUser extends SignInType("current")
 case object GuestUser extends SignInType("guest")
 case object NewUser extends SignInType("new")
```
jfsoul: can't this be taken from the identity API model? also we're sticking with this way of doing user type instead of hasPassword?

app/com/gu/identity/frontend/controllers/SigninAction.scala
Calum-Campbell: True

app/com/gu/identity/frontend/controllers/SigninAction.scala
Calum-Campbell: Its used in a few places in the method, so saves request. in a load of places?

app/com/gu/identity/frontend/logging/MetricsActor.scala
Calum-Campbell: Done

app/com/gu/identity/frontend/models/SignInType.scala
Calum-Campbell: This is what we have in the API at the moment

jfsoul: I mean literally:

`import com.gu.identity.api.model.{GuestUser, NewUser, CurrentUser}`

app/com/gu/identity/frontend/authentication/CookieService.scala
```diff
@@ -59,4 +59,17 @@ object CookieService {
...
+    Seq(PlayCookie(
+      name = "GU_SIGNIN_EMAIL",
+      value = email,
+      maxAge = Some(120),
```
jfsoul: I know this is to pass it along to further stages of the journey - but given we don't know how long people will take it really feels like this should be in a query string parameter rather than a cookie?

app/com/gu/identity/frontend/models/SignInType.scala
Calum-Campbell: Ah this is actually in the identity-api-repo not the model...

jfsoul: Let's move that - it doesn't belong in the repo since we don't store it at all!

Calum-Campbell: ok, I'll do that in the next PR if that's ok

jfsoul: ok sure

app/com/gu/identity/frontend/controllers/SigninAction.scala
```diff
@@ -63,6 +67,31 @@ class SigninAction(
     signInAction(successfulSmartLockSignInResponse, _ => metricsActor.logSuccessfulSmartLockSignin())
   }
 
+  def signInWithEmail = SignInServiceAction(bodyParser) {
+    emailSignInFirstStepAction(successfulFirstStepResponse)
```
walaura: could the naming of these methods be consistent-er?

app/com/gu/identity/frontend/controllers/SigninAction.scala
```diff
@@ -115,6 +143,12 @@ class SigninAction(
     SeeOther(successfulReturnUrl.url)
       .withCookies(cookies: _*)
 
+  def successfulFirstStepResponse(userType: String, successfulReturnUrl: ReturnUrl, cookies: Seq[Cookie], skipConfirmation: Option[Boolean], clientId: Option[ClientID], group: Option[GroupCode]): Result ={
```
walaura: maybe on a separate PR would be helpful to have a signInProperties object to hold all these random form fields such as `skipConfirmation` and `clientId` etc

public/components/two-step-signin/two-step-signin-email.hbs
```diff
@@ -6,6 +6,14 @@
     </div>
   </fieldset>
 
+  {{# group }}
```
walaura: can you move this into `components/two-step-signin/_helpers `?

app/com/gu/identity/frontend/models/SignInType.scala
Calum-Campbell: @jfsoul https://github.com/guardian/identity-frontend/pull/331

PR #324, opened 2018-03-12
Implement smartlock for 2-step signin

walaura:
Missed this?
screen shot 2018-03-12 at 6 15 21 pm

This new implementation shaves off a couple lines of code (namely, it won't try to save user/pwd and trusts the browser to do so. Will have to keep an eye on this in GA and Sentry). No worries though, the old implementation is still in use in the existing signin experience.


public/components/smartlock-trigger/smartlock-trigger.js
```diff
@@ -0,0 +1,77 @@
...
+        // $FlowFixMe
+        if (c instanceof PasswordCredential) {
+          smartLockSignIn(c, returnUrl, csrfToken);
+        } else {
+          throw new Error(ERR_WRONG_CREDENTIAL);
+        }
+      })
+      .catch(err => {
+        throw err;
+      });
```
jfsoul: Is this different to just leaving out the catch? Wouldn't that throw by default?

walaura: it's not!

PR #326, opened 2018-03-13
Robust-er JS code

walaura:
Couple of JS low level improvements:

- Wrap the (new+analytics) js in dotcom's `polyfill.io`, whereas before the js would just not load in older browsers. This makes it run even in IE10
- Wrap new elements in `loadElements` in a `try {} catch {]` block that reports back to Sentry so one error can't bring the rest of components down.


public/js/utils.js
```diff
@@ -0,0 +1,22 @@
...
+    const errorTimeout = window.setTimeout(() => {
+      error();
+    }, 5000);
+    const $script = document.createElement('script');
+    $script.addEventListener('load', () => {
+      clearTimeout(errorTimeout);
+      success();
+    });
+    $script.src =
+      'https://assets.guim.co.uk/polyfill.io/v2/polyfill.min.js?rum=0&features=es6,es7,es2017,Object.assign,Element.prototype.dataset,default-3.6,fetch,promise&flags=gated&callback=onPolyfillIo&unknown=polyfill';
```
sndrs: there's a lot of features being requested here, do you only need `fetch` and `promise`?

you're also passing `callback=onPolyfillIo` which means you can define a global `onPolyfillIo` function to fire your success – `load` doesn't necessarily mean parsed adn ready

walaura: @sndrs i copypasted dotcom and added fetch and promise, i did a very naive check but my idea is that if the browser ~has~ fetch and promise everything else goes too. I'm intentionally casting a wide feature net because in this project want the js to either work or fail but not be in an inconsistent semi working state. i will however remove the more esoteric ones!

nice catch on the callback!

sndrs: in that case, you could just _always_ use it? you always get the callback, and if there's nothing needing to be polyfilled, you just move straight on? 

there will be things that fall through the gaps otherwise, e.g. [ff 45/chrome 49 support `promise` and `fetch` but not `Object.prototype.values`](https://caniuse.com/#feat=object-values,promises,fetch). with the current set up, IE9 (0.01% of guardian visitors) would be polyfilled and work, but the others (0.75%) would error?

PR #328, opened 2018-03-14
Load polyfill and main js in parallel

walaura:
Changes all the scripts on the footer (at this moment that meaning polyfill.io and the main bundle and nothing else ever, everything else should be loaded via better methods) to use `async` to download in parallel and `defer` to run in order. 

This behaviour changes things a bit but not really:

- In browsers that don't support `async` or `defer` nothing changes
- In browsers that support `async` but not `defer` (that's IE9 and lower) (0.08% of id users btw ☹️) The polyfill might load before and break or it might work, this is a 50-50 chance so really you can look at it as affecting 0.04% of users (on a serious note all id-frontend pages work fine without javascript and they are required to work fine without javascript)
- In pretty much every browser both scripts will now download in parallel, then wait for the page to finish parsing (they were already at the footer so no huge change) then run one after each other, ensuring the polyfills are all there

## Screenshots

screen shot 2018-03-14 at 6 15 23 pm <br>Before: Awful, unpolite, look at that bundle tailgating the poor polyfill

screen shot 2018-03-14 at 6 14 57 pm <br>After: The bundle and polyfill are now friends and go hand in hand, and even meet the css in the way!

### DIFF
--- a/public/layout.hbs
+++ b/public/layout.hbs
@@ -45,7 +45,7 @@
   {{#each layout.resources}}
     {{#isJavascript}}
       {{^isInHead}}
-        <script src="{{ url }}" type="text/javascript" async defer></script>
+        <script src="{{ url }}" type="text/javascript" defer></script>
       {{/isInHead}}
     {{/isJavascript}}
   {{/each}}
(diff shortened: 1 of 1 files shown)


walaura:
oops turns out this is useless they already download in parallel!

PR #334, opened 2018-03-21
Send error code -> error text mapping to javascript

walaura:
This PR sends over to the client the error messages for all possible errors, this will let the 2 step sign in javascript flow pick them up and display them without an extra pageload

There's also a helper method that takes a url as a string and returns an array of errors found within it. This is all prep for the 2 step signin js code.

screen shot 2018-03-21 at 1 10 35 pm

### DIFF
--- a/app/com/gu/identity/frontend/views/models/LayoutViewModel.scala
+++ b/app/com/gu/identity/frontend/views/models/LayoutViewModel.scala
@@ -115,7 +115,7 @@ object LayoutViewModel {
         "twoStepSignIn" -> routes.Application.twoStepSignIn().url,
         "smartlockSignIn" -> routes.SigninAction.signInWithSmartLock().url
       ),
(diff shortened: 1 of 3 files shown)


app/com/gu/identity/frontend/views/models/LayoutViewModel.scala
```diff
@@ -114,6 +115,7 @@ object LayoutViewModel {
         "twoStepSignIn" -> routes.Application.twoStepSignIn().url,
         "smartlockSignIn" -> routes.SigninAction.signInWithSmartLock().url
       ),
+      localisedErrors = ErrorViewModel.errorMessages.map{t => (t._1.key, t._2)},
```
mario-galic: You could write `ErrorViewModel.errorMessages.map{case (errId, errMsg) => (errId.key, errMsg)}` if you find this clearer.

walaura: agreed, current solution is a but too clever

PR #337, opened 2018-03-22
Track email inputs in GA

walaura:
Send a `SigninFirstStepSuccessful` to GA when an email is entered in the first field of two step signup. we can use this to track funnels and see how many people drop out after that

### DIFF
--- a/app/com/gu/identity/frontend/analytics/client/MeasurementProtocolRequest.scala
+++ b/app/com/gu/identity/frontend/analytics/client/MeasurementProtocolRequest.scala
@@ -69,6 +69,18 @@ case class SigninEventRequest(request: Request[SignInActionRequestBody], gaUID: 
   override val body = SigninEventRequestBody(request, gaUID)
 }
 
+private object SigninFirstStepEventRequestBody extends MeasurementProtocolRequestBody[SignInActionRequestBody] {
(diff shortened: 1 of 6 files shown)


app/com/gu/identity/frontend/analytics/client/MeasurementProtocolRequest.scala
```diff
@@ -69,6 +69,18 @@ case class SigninEventRequest(request: Request[SignInActionRequestBody], gaUID:
   override val body = SigninEventRequestBody(request, gaUID)
 }
 
+private object SigninFirstStepEventRequestBody extends MeasurementProtocolRequestBody[SignInActionRequestBody] {
+  override val extraBodyParams = Seq(
+    "ea" -> "SigninFirstStepSuccessful",
+    "el" -> "RegularSignin",
```
jfsoul: Should this be `TwoStepSignin` so we can differentiate? Not sure it's super important for this first step, but if we use the same `el` for the actual signin then we can see which ones are from two step?

walaura: this would require to change the events for signinsuccessful too (to an event for the 2 step flow) and at the moment that isn't trivial because normal signin and 2 step signin use the same endpoint.

PR #338, opened 2018-03-22
Break up emails in better places

walaura:
screen shot 2018-03-22 at 1 38 31 pm

When an email address can't fit on the screen, try to break it up at `@`, `+`, and `.` characters first

### DIFF
--- a/public/two-step-signin-choices-page.hbs
+++ b/public/two-step-signin-choices-page.hbs
@@ -8,7 +8,7 @@
       <a class="layout-header" href="{{signinUrl}}" >
         <h1 class="layout-header__title">{{ twoStepSignInPageText.welcome }}</h1>
         <div class="layout-header__title layout-header__title--has-proxy layout-header__title--standfirst">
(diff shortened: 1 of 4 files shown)


public/two-step-signin-choices-page.hbs
```diff
@@ -8,7 +8,7 @@
       <a class="layout-header" href="{{signinUrl}}" >
         <h1 class="layout-header__title">{{ twoStepSignInPageText.welcome }}</h1>
         <div class="layout-header__title layout-header__title--has-proxy layout-header__title--standfirst">
-          <div>{{email}}</div>
+          <div>{{{email}}}</div>
```
mario-galic: Do you need triple curly braces `{{{`?

walaura: yep, i'm sending html code `<wbr>` that would otherwise be escaped by handlebars. This is also why i'm escaping the email address manually back in the viewmodel now!

PR #339, opened 2018-03-22
Fix javascript in 2 step sign in

walaura:
Makes the js handing the 2 step sign in more robust, as well as makes it work at all, since it was temporarily disabled during the endpoint testing

### DIFF
--- a/public/components/two-step-signin/two-step-signin-existing.hbs
+++ b/public/components/two-step-signin/two-step-signin-existing.hbs
@@ -5,7 +5,7 @@
     <div class="form-field-wrap">
       <input class="u-h" name="email" autocomplete="username" value="{{email}}" />
       <label class="form-field-wrap__title" for="signin_field_email">{{twoStepSignInPageText.passwordFieldTitle}}</label>
-      <input class="form-input form-field-wrap__field" type="password" name="password" autocomplete="password" autocapitalize="off" autocorrect="off" spellcheck="false" required="required" autofocus value="{{password}}" />
+      <input class="form-input form-field-wrap__field" type="password" name="password" autocomplete="password" autocapitalize="off" autocorrect="off" spellcheck="false" autofocus value="{{password}}" />
       <div class="u-flexrow form-field-wrap__footer">
         <label class="form-checkbox">
           <input class="form-checkbox__input" type="checkbox" name="rememberMe" checked="checked" value="true"/>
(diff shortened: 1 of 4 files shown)


public/components/two-step-signin/two-step-signin.js
```diff
@@ -94,25 +105,36 @@ const getSlide = ($component: HTMLElement) => {
...
         throw new Error(ERR_MALFORMED_HTML);
       }
+
+      /* naively attempt to preserve password */
+      const $passwordOld = $slide.querySelector('input[name=password]');
+      const $passwordNew = $new.querySelector('input[name=password]');
+      if ($passwordOld && $passwordNew && $passwordNew.parentElement) {
+        $passwordOld.className = $passwordNew.className;
+        $passwordNew.parentElement.replaceChild($passwordOld, $passwordNew);
+      }
```
jfsoul: this seems quite a roundabout way of doing this - are there other options?

public/components/two-step-signin/two-step-signin-existing.hbs
```diff
@@ -5,7 +5,7 @@
     <div class="form-field-wrap">
       <input class="u-h" name="email" autocomplete="username" value="{{email}}" />
       <label class="form-field-wrap__title" for="signin_field_email">{{twoStepSignInPageText.passwordFieldTitle}}</label>
-      <input class="form-input form-field-wrap__field" type="password" name="password" autocomplete="password" autocapitalize="off" autocorrect="off" spellcheck="false" required="required" autofocus value="{{password}}" />
+      <input class="form-input form-field-wrap__field" type="password" name="password" autocomplete="password" autocapitalize="off" autocorrect="off" spellcheck="false" autofocus value="{{password}}" />
```
jfsoul: how come this no longer needs to be required?

public/components/two-step-signin/two-step-signin.js
walaura: we are hacking our way around browser password managers so there's no 'clean' way to do this. Same goes for password manager extensions! They all hack their own way around forms. I figured literally moving the password input around fields is the safest way to do this and once you build on top of that this is the most 'reusable' solution. as in, hypothetically all flows where after typing an email there's a password would work seamlessly, back and forth

The actual 'clean' way to do this is to use smartlock instead of presenting a password field at all, which we do! so in reality this is just to get a nice autocompleted password field

public/components/two-step-signin/two-step-signin-existing.hbs
walaura: technically it does, however, because of the password manager solution, this field can end up on the first page (enter email, go to password screen, change email) which would render you unable to continue, as the field is required but invisible.

There isn't really anything wrong resulting of not requiring the field. it lets you submit an empty password but it won't really do much, just fail which would be expected.

PR #342, opened 2018-03-22
Report errors from 2 step sign in JS

walaura:
Report back errors gotten while signing in via js. This PR mostly adds the scaffolding to then report errors to.

screen shot 2018-03-29 at 2 55 11 pm


rahilb:
what does the compatibility mode link do?

walaura:
@rahilb not a thing for now! it's an upcoming-ish feature that will let users load the site with javascript disabled just in case.

PR #343, opened 2018-03-23
Clean up js

walaura:
Did a round of browserstacking and solved a nasty animation-related bug + refactored `two-step-signin__slide.js`

### DIFF
--- a/public/components/two-step-signin/two-step-signin.js
+++ b/public/components/two-step-signin/two-step-signin.js
@@ -23,35 +23,44 @@ const pushSlide = (
       }
     : { in: 'two-step-signin__slide--in', out: 'two-step-signin__slide--out' };
 
-  return new Promise(resolve => {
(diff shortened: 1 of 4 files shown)


public/components/two-step-signin/two-step-signin.js
```diff
@@ -23,35 +23,44 @@ const pushSlide = (
...
+            $new.classList.add(_)
+          );
+        });
+      } else {
+        $old.remove();
+        resolve($new);
+      }
     });
 
+  return Promise.resolve().then(() => {
```
jfsoul: Is this a standard pattern? Could you not just create a new promise to do the same thing?

walaura: it's a pattern for when you have a long promise chain so they all look more alike vs a having a normal promise constructor then a promise chain but in this case there isn't really any point to it. good catch!

PR #346, opened 2018-03-26
Larger sign up button

walaura:
screen shot 2018-03-26 at 3 46 02 pm

Not a fan of the visual design but it is larger and more prominent

@katyabeer

### DIFF
--- a/conf/messages.en
+++ b/conf/messages.en
@@ -13,8 +13,8 @@ signin.password=Password
 signin.forgottenpassword=Set or reset your password
 signin.rememberme=Remember me
 signin.signin=Sign in
-signin.noaccount=Don''t have an account?
-signin.signup=Sign up
+signin.noaccount=New to The Guardian?
+signin.signup=Create an account
 signin.continue=Continue
 
 signinTwoStep.welcome=Welcome
--- a/public/_html.css
+++ b/public/_html.css
@@ -44,8 +44,7 @@ body > section:last-of-type {
   padding: 0 1rem;
 
   @media (--viewport-min-mobile-landscape) {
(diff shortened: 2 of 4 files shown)


katyabeer:
It's hoooooooge! :)

katyabeer:
A couple of things: 
- Copy: "New to the Guardian?"
- Button copy: 'Create an account'.
- Can we shorten the button somehow? :) At the moment it's quite a sausage!

jfsoul:
Putting my paranoid hat on - the new copy `New to The Guardian?` might not be applicable in all cases.  People can be very familiar with the Guardian - even have a paying relationship without having an account.

Also @katyabeer what was wrong with `Sign up`? `Create an account` sounds so formal.

katyabeer:
@jfsoul 

"New to the Guardian?" - Good point. Will double-check the wording with Celine asap.

'Create an account':

- It's not good practice to have ‘Sign In’ and ‘Sign Up’ on the same page/as a combination. When buttons/CTAs both use the same verb in their labels it’s easy for users to get confused. Even if users didn’t make the mistake of clicking on the wrong button, they’ll spend extra time distinguishing the two. So we either use Sign in / Register. Or Sign in / Create an account. Since the Sign up page is actually called 'Create your Guardian account' I'd keep it consistent... https://profile.theguardian.com/register?returnUrl=https%3A%2F%2Fwww.theguardian.com%2Fuk

katyabeer:
I've just spoken to Celine. She's fine with 'New to the Guardian?' because of the context that line appears in. It is implied that on the 'Sign page' the context is about having an account and **being** a **new** (or returning) user.

PR #348, opened 2018-03-27
Improve GA tracking

walaura:
* Track ajax pv's in GA
* Support `data-link-name` for sending events (to GA but not DL)


walaura:
oh man i always thought this was mario personally verifying each commit

PR #351, opened 2018-03-29
Password sign in via javascript

walaura:
Sign in (2 step) with a password via javascript too, making the flow neater and more lively.

screen shot 2018-03-29 at 4 04 29 pm

### DIFF
--- a/app/com/gu/identity/frontend/controllers/SigninAction.scala
+++ b/app/com/gu/identity/frontend/controllers/SigninAction.scala
@@ -80,7 +80,7 @@ class SigninAction(
   }
 
   def signInSecondStepCurrent = signInSecondStepCurrentServiceAction(bodyParser) {
-    signInAction(successfulSignInResponse, signInMetricsLogger)
+    signInAction(successfulSignInResponse, successfulAjaxSignInResponse, signInMetricsLogger)
   }
 
   def signIn = SignInServiceAction(bodyParser) {
--- a/app/com/gu/identity/frontend/controllers/SigninAction.scala
+++ b/app/com/gu/identity/frontend/controllers/SigninAction.scala
@@ -84,11 +84,11 @@ class SigninAction(
   }
 
   def signIn = SignInServiceAction(bodyParser) {
-    signInAction(successfulSignInResponse, signInMetricsLogger)
+    signInAction(successfulSignInResponse, successfulAjaxSignInResponse, signInMetricsLogger)
   }
 
   def signInWithSmartLock = SignInSmartLockServiceAction(bodyParser) {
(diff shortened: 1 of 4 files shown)


Calum-Campbell:
I think this needs to change for the new post endpoint on this form

public/components/two-step-signin/two-step-signin__slide.js
```diff
@@ -10,10 +11,14 @@ const SLIDE_STATE_DEFAULT: string = 'SLIDE_STATE_DEFAULT';
 
 const EV_DONE: string = 'form-done';
 
-const ERR_MALFORMED_HTML: string = 'Something went wrong';
-const ERR_MALFORMED_RESPONSE: string = 'Something went wrong';
+const ERR_MALFORMED_HTML: string = 'Something went wrong (html)';
```
Calum-Campbell: I would prefer this to say "malformed html", but it doens't seem like it's used anyway so maybe can get rid?

walaura: yeah, the error reporting is a bit of a mess but ideally this _may_ end up being user facing, for now this is mostly so we ca pick them up on sentry

app/com/gu/identity/frontend/controllers/SigninAction.scala
```diff
@@ -134,7 +134,12 @@ class SigninAction(
       case Right(cookies) => Right {
         if (stage == "PROD") Tip.verify("Account Signin")
         metricsLogger(request)
-        successResponse(successfulReturnUrl, cookies)
+        if(request.headers.toSimpleMap.contains("x-gu-browser-rq")){
+          successAjaxResponse(successfulReturnUrl, cookies)
+        }
```
rahilb: the formatting is wrong, it should be 

`} else {`

app/com/gu/identity/frontend/controllers/SigninAction.scala
```diff
@@ -168,6 +173,12 @@ class SigninAction(
     SeeOther(successfulReturnUrl.url)
       .withCookies(cookies: _*)
 
+
+  def successfulAjaxSignInResponse(successfulReturnUrl: ReturnUrl, cookies: Seq[Cookie]): Result =
+    Ok("{\"status\": true, \"returnUrl\": \"" + successfulReturnUrl.url + "\"}")
```
rahilb: you can use triple quotes `"""{"foo":"bar"}"""` and then you don't need to do this escaping.

rahilb: Aditionally, this is prone to error if we have bad values (won't be escaped)... we should always construct json in a typesafe way using the AST.

PR #352, opened 2018-04-03
New endpoint and redirect for existing users to post password to

Calum-Campbell:
- When an existing user gets to the second step in the two step sign in , this posts to an endpoint that will redirect back to the second step (/signin/current) if the password is wrong or there is an error. Previously this would result in a redirect to /signin.

### DIFF
--- a/app/com/gu/identity/frontend/controllers/SigninAction.scala
+++ b/app/com/gu/identity/frontend/controllers/SigninAction.scala
@@ -17,6 +17,8 @@ import com.gu.identity.frontend.authentication.CookieService
 import com.gu.tip.Tip
 import java.net.URLEncoder.encode
 
+import com.gu.identity.model.CurrentUser
+
 import scala.concurrent.Future
 
 /**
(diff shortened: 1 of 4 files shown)


app/com/gu/identity/frontend/controllers/SigninAction.scala
```diff
@@ -35,12 +35,20 @@ class SigninAction(
 
   val redirectRoute: String = routes.Application.signIn().url
 
+  val signInSecondStepCurrentRedirectRoute: String = routes.Application.twoStepSignInStepTwo("current").url
```
walaura: not a fan of "current" being a string, maybe this should be a case class?

Calum-Campbell: Good shout, using CurrentUser.name now.

walaura:
[/actions/signin should return to /signin/{signinType} on error](https://trello.com/c/ue2L662E/433-actions-signin-should-return-to-signin-signintype-on-error)

PR #355, opened 2018-04-06
Magic link exchange

rahilb:
frontend endpoints for magic link signin

### DIFF
--- a/app/com/gu/identity/frontend/controllers/SigninAction.scala
+++ b/app/com/gu/identity/frontend/controllers/SigninAction.scala
@@ -172,8 +172,8 @@ class SigninAction(
     identityService.sendSignInTokenEmail(req, ClientIp(_req)).map {
       case Right(_) =>
         SeeOther(routes.Application.sendSignInLinkSent().url)
-      case Left(_) =>
-        renderErrorPage(config, BadRequestError("Unable to send SignIn link."), Results.BadRequest.apply)
+      case Left(errors) =>
+        SeeOther(routes.Application.sendSignInLink(error = errors.map(_.getMessage)).url)
     }
   }
 
--- a/app/com/gu/identity/frontend/controllers/SigninAction.scala
+++ b/app/com/gu/identity/frontend/controllers/SigninAction.scala
@@ -171,7 +171,7 @@ class SigninAction(
     val req = _req.body
     identityService.sendSignInTokenEmail(req, ClientIp(_req)).map {
       case Right(_) =>
-        renderSendSignInLinkSent(config)
+        SeeOther(routes.Application.sendSignInLinkSent().url)
       case Left(_) =>
         renderErrorPage(config, BadRequestError("Unable to send SignIn link."), Results.BadRequest.apply)
     }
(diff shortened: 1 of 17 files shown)


app/com/gu/identity/frontend/controllers/SigninTokenController.scala
```diff
@@ -0,0 +1,29 @@
...
+  val messagesApi: MessagesApi,
+  implicit val executionContext: ExecutionContext
+) extends Controller with LazyLogging with I18nSupport {
+
+  def signinWithToken(token: String, returnUrl: Option[String]): Action[AnyContent] = Action.async {
+    identityService.authenticateSigninToken(token).map {
+      case Right(cookies) =>
+        SeeOther(returnUrl.getOrElse(configuration.dotcomBaseUrl)).withCookies(cookies: _*)
+      case Left(_) =>
+        ViewRenderer.renderErrorPage(configuration, SigninTokenRejected("The sign in token was expired or invalid, please request a new one."), Results.Unauthorized.apply)
```
walaura: could you use the localised error text here instead of a string?

conf/routes
```diff
@@ -19,6 +19,7 @@ GET        /signin-token                                com.gu.identity.frontend
 
 GET        /signin-token/email-sent                     com.gu.identity.frontend.controllers.Application.sendSignInLinkSent
 
+POST       /signin-token/email                          com.gu.identity.frontend.controllers.SigninAction.sendSignInLink()
```
walaura: as per convention this needs to begin with `/actions/`

app/com/gu/identity/frontend/controllers/SigninAction.scala
```diff
@@ -169,6 +167,16 @@ class SigninAction(
     }
   }
 
+  def sendSignInLink(): Action[EmailSignInRequest] = Action.async(emailSigninFormParser) { _req =>
+    val req = _req.body
+    identityService.sendSignInTokenEmail(req, ClientIp(_req)).map {
+      case Right(_) =>
+        renderSendSignInLinkSent(config)
```
walaura: rendering pages from POST is a bit of a mess, could you turn this into a 303 redirect?

rahilb: Why? This is how we end up with 100s of requests to load a single page... we know what we want to do at this point so why not just do it?

conf/routes
rahilb: I'll make the change, but I don't think this is a good convention: Are we creating an `action`? semantically the API does not make sense.

app/com/gu/identity/frontend/controllers/SigninAction.scala
rahilb: Additionally, we've already had bugs caused by redirects from post requests...

rahilb: OK, what should the redirect be for the error? Or should I just render it?

rahilb: BTW the bug i was mainly referring to was posting of consents (that was returning a 303 that was not checked by frontend code).

walaura: Normally in this project the errors go back to the previous route with a `?error` parameter like this: `https://profile.theguardian.com/signin?error=signin-error-credentials` (I believe it's comma separated but i've only ever seen it return one at a time) and then the preceding route picks it up!

Regarding the error this ties into our previous conversation about the difference between a POST endpoint on an API (should be pure data) and a basic HTML POST form like this one which should return users to the proper route

rahilb: right, so endpoints should all know their own call sites so we can use http redirects for control flow, that's insane. 

I'll make the change to unblock you but i think it is ridiculous.

PR #356, opened 2018-04-06
Move the ajax form bits of 2 step signin into .ajax-step-flow

walaura:
`ajax-step-flow` will also be used in the request token page so this decouples that from two step signin into its own component. This is a major win for code readability too as everything is now more neatly separated.

The changes are mostly renaming but there's some areas of note:

- Removed `two-step-signin-container` and `two-step-signin-form_section` as they were unused
- Removed `components/two-step-signin/scriptFile.scala` which was an unused editor mistake
- Refactored the back link behaviour in `ajax-step-flow__slide` so it's more reusable instead of a direct mention to the reset link


public/components/ajax-form/ajax-form__slide.js
```diff
@@ -96,7 +104,7 @@ const fetchSlide = (action, $stateable, fetchProps) =>
       });
     });
 
-const catchSlide = ($stateable, err) => {
```
jfsoul: `stateable` doesn't mean much to me.  What is it supposed to be?

public/components/ajax-form/ajax-form__slide.js
```diff
@@ -109,52 +117,62 @@ const catchSlide = ($stateable, err) => {
...
+    .filter(_ =>
+      validAjaxLinkRoutes.map(r => _.href.contains(r)).some(c => c === true)
+    );
+
+  const $forms: HTMLFormElement[] = [
+    ...($component.querySelectorAll(`form`): any)
+  ]
+    .filter(_ => _ instanceof HTMLFormElement)
+    .filter(_ =>
+      validAjaxFormRoutes.map(r => _.action.contains(r)).some(c => c === true)
```
jfsoul: is the same as `validAjaxFormRoutes.some(r => _.action.contains(r))`?

walaura:
@jfsoul it really aims to be a pluggable thing you can do to any form. In reality this will be used for signin, potentially signup, and the magic link sender tool (which will become part of the signin flow) so i'm happy to rename it but i'm not sure what to

public/components/ajax-form/ajax-form__slide.js
walaura: nice catch!

public/components/ajax-form/ajax-form__slide.js
walaura: i don't know either which is a very telling sign of a bad name, i'll check on it

jfsoul:
Re: `ajax-form` - let's leave it as is for now, but maybe think about putting in a short comment at the top of the file explaining what it's for

PR #358, opened 2018-04-10
Turn email field into a button in 2ss

walaura:
For the UX test. Nothing super fancy going on but it works!

screen shot 2018-04-10 at 4 25 55 pm
screen shot 2018-04-10 at 4 26 00 pm


walaura:
@jfsoul Grabbed it from gu.com That said if we go with this direction we definitely need a visual designer to check up on us, the current state is a bit unengaging on desktop.

PR #359, opened 2018-04-11
Add form for the MLST (magic link sender tool)

walaura:
I believe once this is merged we can begin to use it live!

Should we be collecting any GA metrics? might be useful.

screen shot 2018-04-11 at 11 07 21 am
screen shot 2018-04-11 at 11 07 15 am


walaura:
aand here we go!

PR #367, opened 2018-04-17
Rename signin token stuff to resub-email

rahilb:
Renames all routes related to `signin-token` to `resub-email` to reflect what the original requirements intended. 

related PR https://github.com/guardian/identity/pull/1097


app/com/gu/identity/frontend/views/models/LayoutViewModel.scala
```diff
@@ -116,7 +116,7 @@ object LayoutViewModel {
         "twoStepSignIn" -> routes.Application.twoStepSignInStart().url,
         "twoStepSignInAction" -> routes.SigninAction.emailSignInFirstStep().url,
         "signInSecondStepCurrentAction" -> routes.SigninAction.signInSecondStepCurrent().url,
-        "sendSignInLinkAction" -> routes.SigninAction.sendSignInLinkAction().url,
+        "sendSignInLinkAction" -> routes.SigninAction.sendResubLinkAction().url,
```
rahilb: @walaura  none of us could figure out where this string was used so I didn't change it.

jfsoul: Ah I see now - the routes (as above) are used in the javascript and added to the page via direct insertion into the html, but the specific view models each have an `actions` property as required for the handlebars templates.

walaura: This has nothing to do with handlebars. These routes are for javascript to pick up directly [like this](https://github.com/guardian/identity-frontend/blob/a5d91c548462add7d0cd406886cbb0871a4df911/public/components/ajax-step-flow/_valid-routes.js#L6). It's concerning that IJ wasn't able to find this usage though because it's in master at the moment and there's no automated way to test for these if they break.

PR #373, opened 2018-04-20
Update TOS text

walaura:
screen shot 2018-04-20 at 12 21 30 pm
screen shot 2018-04-20 at 12 21 33 pm

Not a huge fan of partner text, that could be a different paragraph probably?

### DIFF
--- a/app/com/gu/identity/frontend/models/text/TermsText.scala
+++ b/app/com/gu/identity/frontend/models/text/TermsText.scala
@@ -4,6 +4,7 @@ import play.api.i18n.Messages
 
 case class TermsText private(
      conditionsText: String,
+     dataText: String,
      termsOfServiceText: String,
      termsOfServiceUrl: String,
      privacyPolicyText: String,
@@ -21,6 +22,7 @@ object TermsText {
   def apply()(implicit messages: Messages): TermsText =
     TermsText(
       conditionsText = messages("terms.conditions"),
+      dataText = messages("terms.dataConditions"),
       termsOfServiceText = messages("terms.termsOfService"),
       termsOfServiceUrl = messages("terms.termsOfServiceUrl"),
       privacyPolicyText = messages("terms.privacyPolicy"),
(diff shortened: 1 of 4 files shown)


jfsoul:
image
I think one of the reasons we are changing this wording is that you can't agree to a privacy policy, so this wording probably needs work still.

"By proceeding, you agree to our Terms of Service and Guardian Jobs' Terms of Service.  To find out what personal data we collect and how we use it, please visit our Privacy Policy and Guardian Jobs' Privacy Policy".

Also will this work for teacher network too?

walaura:
All set! I refactored it all and tested both partners/groups but would like a second set of eyes since this is very legal very serious stuff

app/com/gu/identity/frontend/models/text/TermsText.scala
```diff
@@ -3,43 +3,61 @@ package com.gu.identity.frontend.models.text
...
+      messages("terms.privacyPolicyUrl")
+    )
     TermsText(
-      conditionsText = messages("terms.conditions"),
-      termsOfServiceText = messages("terms.termsOfService"),
-      termsOfServiceUrl = messages("terms.termsOfServiceUrl"),
-      privacyPolicyText = messages("terms.privacyPolicy"),
-      privacyPolicyUrl = messages("terms.privacyPolicyUrl")
+      conditionsText = messages("terms.conditions",conditionsLink),
+      privacyText = messages("terms.privacy",privacyLink)
```
jfsoul: add spaces after commas here

app/com/gu/identity/frontend/views/models/TermsViewModel.scala
```diff
@@ -1,93 +1,60 @@
...
 case class BasicTermsViewModel private(
-    conditionsText: String,
-    termsOfServiceText: String,
-    termsOfServiceUrl: String,
-    privacyPolicyText: String,
-    privacyPolicyUrl: String)
-  extends ViewModel with TermsViewModel {
-}
+  conditionsText: String,
+  privacyText: String) extends ViewModel with TermsViewModel {}
```
jfsoul: I don't think you should need the curly braces after these?

walaura:
@jfsoul should we with Group as that's what the query string is called?

jfsoul:
If you really want my opinion then I never found `group` a very helpful name.  But not a big deal either way if you have a strong opinion.

walaura:
@jfsoul i don't like `group` either but I feel it's clearer to have 1 universal name for the concept than have 2 and know they are connected and I feel that's clearer

PR #374, opened 2018-04-23
Add campaign code to resub email redirect

rahilb:
Adds a campaign code to resub-email consents redirects as mentioned in a standup last week.


rahilb:
Not too sure about INTCMP vs CMP, I got these values from @walaura 🙈

walaura:
no idea :( these aren't standard tags, as far as I know, maybe the INT stands for international?

PR #379, opened 2018-04-25
Add webmail links to resub-email/email-sent

walaura:
screen shot 2018-04-25 at 5 07 59 pm

### DIFF
--- a/app/com/gu/identity/frontend/models/EmailProvider.scala
+++ b/app/com/gu/identity/frontend/models/EmailProvider.scala
@@ -1,0 +1,48 @@
+package com.gu.identity.frontend.models
+
+/**
+ * Defines a SignInType for determining where to point back links to.
+ */
+
+sealed trait EmailProvider extends Product2[EmailProvider, String] {
+  self =>
+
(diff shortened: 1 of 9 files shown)


app/com/gu/identity/frontend/models/EmailProvider.scala
```diff
@@ -0,0 +1,48 @@
+package com.gu.identity.frontend.models
+
+/**
+ * Defines a SignInType for determining where to point back links to.
+ */
+
+sealed trait EmailProvider extends Product2[EmailProvider, String] {
```
rahilb: You probably don't need to extend Product2, and you can delete the `def _1` & `def _2`. Then instead of getting id by calling  `._2` you can just use `.id`

app/com/gu/identity/frontend/models/EmailProvider.scala
```diff
@@ -0,0 +1,48 @@
...
+  self =>
+
+  val id: String
```
rahilb: you can make these defs, and just override with `vals`

PR #382, opened 2018-04-27
Enable global CSRF filter

mario-galic:
https://www.playframework.com/documentation/2.5.x/ScalaCsrf#Applying-a-global-CSRF-filter

* Automatically add and check CSRF tokens to requests.
* I have tested by fiddling with the token and Play did indeed detect it.


walaura:
Haven't had a look but this can potentially break the guardian.com/staywithus landing page which can't send a token. 

I would personally argue against having tokens at all on this project as these are public facing pages. But if we are keeping them i would need resub links to be checked to work without them. At the moment it was implemented but never checked, ideally we would remove it altogether in that page

jfsoul:
Bit strange that the landing page posts directly to identity-frontend api.  Is it that hard to implement in dotcom as well?

With regards to the CSRF tokens - there is one non-public page for accepting third party Ts&Cs and hopefully we will add more over time, so this seems reasonable to enable.

mario-galic:
Close in favour off https://github.com/guardian/identity-frontend/pull/384

walaura:
@jfsoul the landing page doesn't live in dotcom but in the interactive repo which has no server-side code. Our options are

– Only make it work with javascript and point directly to the API (clean but less compatible and we need every single request we can)
– Copy the endpoint to dotcom then point it to dotcom (unclean and duplicating code)
– Make a lambda to proxy the api just for this 1 use case (I like this but feels like overengineering)
– Point it to id-front's server handling (unclean but it's 0 extra lines of code)

Going forward I think maybe a good compromise would be to, just like we have `/actions/` as a prefix for POST endpoints have `/external-actions/` as a prefix for non CSRF . This is something we will also need if we ever move the sign in on site for example or if we migrate newsletter subscriptions

PR #387, opened 2018-05-01
Remove custom parser hardwiring

mario-galic:
Addresses `BodyParsers.parse` deprecation warning:

```
@deprecated("Inject PlayBodyParsers or use AbstractController instead", "2.6.0")
```

### DIFF
--- a/app/com/gu/identity/frontend/configuration/FrontendApplicationLoader.scala
+++ b/app/com/gu/identity/frontend/configuration/FrontendApplicationLoader.scala
@@ -10,6 +10,7 @@ import com.gu.identity.frontend.errors.ErrorHandler
 import com.gu.identity.frontend.filters._
 import com.gu.identity.frontend.logging._
 import com.gu.identity.frontend.mvt.MultiVariantTestAction
+import com.gu.identity.frontend.request._
 import com.gu.identity.frontend.services._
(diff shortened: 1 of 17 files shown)


app/com/gu/identity/frontend/configuration/FrontendApplicationLoader.scala
```diff
@@ -53,6 +54,14 @@ class ApplicationComponents(context: Context)
 
   lazy val identityCookieDecoder: IdentityCookieDecoder = new IdentityCookieDecoder(IdentityKeys(frontendConfiguration.identityCookiePublicKey))
 
+  // Parsers
+  lazy val formRequestBodyParser = new FormRequestBodyParser(playBodyParsers)
+  lazy val emailResubRequestsParser = new EmailResubRequestsParser(formRequestBodyParser)
+  lazy val signInActionRequestBodyParser = new SignInActionRequestBodyParser(formRequestBodyParser)
```
rahilb: It looks like these map 1 to 1 with the controllers, can we not just inject `formRequestBodyParser` and construct each one inside the controller?

mario-galic: True, let this PR address warnings and we will cleanup in separate PR.

PR #390, opened 2018-05-08
Enable source maps for Sentry

mario-galic:


sndrs:
not how we do it (we use config) but looks good to me!

PR #397, opened 2018-06-05
Remove opt in branding

walaura:
- Change the redirect in `RepermissionController` to go to `/consents` instead of `/consents/staywithus`

- Remove the opt-in clientID and the "✉️issing something?" banner it displayed:

screen shot 2018-06-05 at 10 23 30 am

I think this is actually all of it! Can't find anything else in the whole project by searching 'opt-in', 'optin', or 'stay'


rahilb:
I am issing something, this incredible banner

PR #401, opened 2018-06-07
Remove custom appender and old ELK, add new central ELK plugin

Calum-Campbell:
- Remove custom appender code for ELK logging
- Add logstash plugin
- Point to new central ELK stack


app/com/gu/identity/frontend/logging/SuccessfulActionCloudwatchLogging.scala
```diff
@@ -23,10 +23,10 @@ object LoggingAsyncHandler extends AsyncHandler[PutMetricDataRequest, PutMetricD
 object SuccessfulActionCloudwatchLogging {
 
   private lazy val cloudwatch = {
-    AmazonCloudWatchAsyncClientBuilder.standard()
```
Calum-Campbell: @aug24  not sure if this needs reverting?

aug24: No, I don't think it should be.  The old way is deprecated.

PR #406, opened 2018-06-12
Fix text contrast

walaura:
Remove `--color-text-light ` and make all text `#121212` instead. The visual changes are barely noticeable but this one passes WAI AAA contrast standards


walaura:
@rahilb we can automate [lighthouse](https://developers.google.com/web/tools/lighthouse/) as part of CI!

PR #407, opened 2018-06-12
Remove dupe smartLockSignIn metric

walaura:
The smartlocksignin event is being sent twice to GA at the moment. This removes the client side sending

### DIFF
--- a/public/components/smartlock-trigger/smartlock-trigger.js
+++ b/public/components/smartlock-trigger/smartlock-trigger.js
@@ -2,7 +2,6 @@
 
 import qs from 'qs';
 import { route } from 'js/config';
-import { customMetric } from 'components/analytics/ga';
 
 const selector: string = '.smartlock-trigger';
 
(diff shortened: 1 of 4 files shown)


public/components/smartlock-trigger/smartlock-trigger.js
```diff
@@ -22,23 +22,27 @@ const smartLockSignIn = (
...
+        body: qs.stringify({
+          email: credentials.id,
+          password: credentials.password,
+          gaClientId: tracker.get('clientId'),
+          csrfToken
+        })
+      })
+    )
+    .then(r => {
+      if (r.status === 200) {
```
rahilb: `r.statusCode.isSuccess`?

rahilb:
still LGTM 🚢

public/components/smartlock-trigger/smartlock-trigger.js
walaura: so proud of you 🙏

PR #415, opened 2018-06-20
Change copy for 'Remember me' on the 1-step signin

walaura:
Blink it and you'll miss it!

Code-wise I refactored some of the css to avoid having `u-flexrow` and `layout-sbs-col` doing almost the same thing 9putting things side by side). THis was mostly to make the new checbox stack on mobile, as the longer copy means it can't fit on the screen otherwise.

/cc @katyabeer

## Before / After
screen shot 2018-06-20 at 12 00 55 pm
screen shot 2018-06-20 at 12 00 53 pm


rahilb:
I actually think this ruins the symmetry of the page.

PR #423, opened 2018-06-21
Remove the gu_digital_subscriber_cookie on signout

lmath:
We want to stop showing reader revenue messaging to digital subscribers. So, we will be adding a gu_digital_subscriber cookie here: https://github.com/guardian/frontend/pull/19897

This cookie should be removed on sign out. 

cc @guardian/identity


JustinPinner:
LGTM 👍 Anyone else?

PR #425, opened 2018-06-25
Bespoke styling for recurring contributors

jranks123:
For those pushed to sign in whilst making a recurring contribution, we want to keep the header consistent with that of [support.theguardian.com](https://support.theguardian.com), and also add some custom copy. 

The change to the header affects any page on identity frontend when the `clientId` is `recurringContributions`. Currently nothing is coming to this page with this `clientId`, but that will change with[ this test ](https://github.com/guardian/support-frontend/pull/739).

| Original sign in page |
| -----------------|
|page1original|

|New page for recurring contributors|
|-----|
|page1recurring|

@guardian/contributions


app/com/gu/identity/frontend/models/Text.scala
```diff
@@ -47,7 +47,7 @@ object Text {
         "continue" -> messages("signin.continue"),
         "termsOfService" -> messages("signin.termsofservice"),
         "privacyPolicy" -> messages("signin.privacypolicy"),
-
+        "supportingParagraph" -> (if(clientId.contains(GuardianRecurringContributionsClientID)) messages("signinTwoStep.recurringSupportingParagraph") else ""),
```
walaura: absolutely not a blocker (In fact I'd advise to make this a separate PR) Could we turn this into an array?

PR #427, opened 2018-06-27
Implement changes for the recurringContributions clientId

jranks123:
When people hit identity with the new `recurringContributions` client ID, we want them to:

**a)** See the new header on the `/register` and `/signin/start` flows
**b)** See the "Support the guardian" title with a standfirst saying "Create your Guardian account" on the `/register` page
**c)**  See the custom title and copy on the `/signin/start` flow

**a)** and **c)** were implemented in [this PR](https://github.com/guardian/identity-frontend/pull/425). This PR implements **b)**, and also sets the default returnUrl to `support.theguardian.com` (although the returnUrl should be passed through from support frontend so hopefully this won't be needed, I'm just following the pattern for the `members` clientId)

In the near future we should stop sending the `members` clientId from support frontend, and create a new, support specific clientId, as currently the default return url for `members` is the membership site. At the moment we are always passing a returnUrl to the identity site so it isn't causing an issue right now but it may well do in the future.

Below I have attatched screenshots showing the effects of **a)**, **b)** and **c)**

| /register before | /register after |
|----------|--------|
|registerbefore|registerafter|

| /signup/start before  | /signup/start after |
|----------|--------|
|2stepbefore|2stepafter|

### DIFF
--- a/app/com/gu/identity/frontend/models/ReturnUrl.scala
+++ b/app/com/gu/identity/frontend/models/ReturnUrl.scala
@@ -48,9 +48,13 @@ object ReturnUrl {
   def defaultForMembership(configuration: Configuration) =
     defaultFromUrl(configuration.preferredMembershipUrl)
 
+  def defaultForSupport(configuration: Configuration) =
+    defaultFromUrl(configuration.preferredSupportUrl)
+
   def defaultForClient(configuration: Configuration, clientId: Option[ClientID]) =
     clientId match {
       case Some(GuardianMembersClientID) => defaultForMembership(configuration)
+      case Some(GuardianRecurringContributionsClientID) => defaultForSupport(configuration)
       case _ => default(configuration)
     }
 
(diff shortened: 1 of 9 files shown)


app/com/gu/identity/frontend/models/ReturnUrl.scala
```diff
@@ -48,8 +48,12 @@ object ReturnUrl {
   def defaultForMembership(configuration: Configuration) =
     defaultFromUrl(configuration.preferredMembershipUrl)
 
+  def defaultForSupport(configuration: Configuration) =
+    defaultFromUrl(configuration.preferredSupportUrl)
+
   def defaultForClient(configuration: Configuration, clientId: Option[ClientID]) =
     clientId match {
+      case Some(GuardianMembersClientID) => defaultForMembership(configuration)
```
walaura: I think this should be `defaultForSupport `?

jranks123: Really, what we need to do is stop sending the Members Client Id as the default from the support page, and create a new one called `support`. But I don't want to change the behaviour of the `members` clientId, as the membership site is still live. Currently on the support site, the only product that sends people to sign in is recurring contributions, so in the meantime it is ok to assume that anyone coming from support will have the `GuardainRecurringContributionsClientId`

walaura: ah!

PR #428, opened 2018-06-27
[WIP] React islands

walaura:
Prep work for Uppsala. The description is a bit too long for a PR so [here's a document](https://docs.google.com/document/d/1ny4CXg1W_i66HgIFhcVZE2JXqftNGCpXjzt6SBmJKsg/edit?usp=sharing) detailing the changes. 

I think this approach overall is pretty good for our purposes and has the least disruption, while still letting us write pretty modern react code from day 1

## Todo 
- [ ] Update docs
- [x] Go from preact to react

island-transparent-png


jfsoul:
As instructed I'm not worrying too much about the code, but I wholeheartedly approve of the principles you outlined in the doc 💯 👍

app/com/gu/identity/frontend/views/models/LayoutViewModel.scala
```diff
@@ -155,7 +155,7 @@ object LayoutViewModel {
       InlinedJSONResource("id_runtime_params", r.toJSONString)
     }
 
-    val cssResources = LocalCSSResource.fromAsset("bundle.css")
+    val cssResources = LocalCSSResource.fromAsset("main.jaffa.css")
```
walaura: I wanted to add a touch of whimsy and character to this but I want everybody to be happy with the name before we commit to it 🍊

.eslintrc.js
```diff
@@ -20,10 +20,13 @@ module.exports = {
     'import/no-extraneous-dependencies': 'off',
     'import/prefer-default-export': 'off',
     'prettier/prettier': 'error',
+    'jsx-quotes': ['error', 'prefer-double'],
     'no-extend-native': 'error',
     'no-param-reassign': ['error', { props: false }],
     'func-style': ['error', 'expression', { allowArrowFunctions: true }],
     'prefer-destructuring': 'off',
+    'react/jsx-filename-extension': 'off',
+    'react/prefer-stateless-function': 'off',
```
RichieAHB: I guess this is just so you don't have to refactor some old classes?

.flow-typed/global.js
```diff
@@ -0,0 +1 @@
+declare var React: $Exports<'react'>;
```
RichieAHB: I think flow has this by default doesn't it? Does it not work without this?

public/components/form/_form-button.css
```diff
@@ -1,3 +1,6 @@
+@import '../../css/_vars.css';
```
RichieAHB: Would it be preferable to add some form of resolve root here to help with refactoring in future, using something like https://github.com/css-modules/postcss-modules-resolve-imports - not a requirement but often it's quite useful 😀

.eslintrc.js
walaura: Ah! this is actually `prettier` bumping heads with jsx. There's a long-standing health task to make linting more decent ToT

.flow-typed/global.js
walaura: This one's a holdover from trying to use preact instead of react! I'll try removing it

public/components/form/_form-button.css
walaura: Absolutely agree, will make those work better!

Thanks for all the comments!

.eslintrc.js
RichieAHB: Merging linting tools is perhaps the worst thing about modern js 😞

app/com/gu/identity/frontend/views/models/LayoutViewModel.scala
jfsoul: I just worry that someday, someone else will come across this and spend time trying to work out what `jaffa` means 😞

PR #443, opened 2018-07-23
More sentry js improvements

walaura:
#441 was a good first start but sadly the hacky way we were reporting errors doesn't cut it anymore so this PR makes them ENTERPRISE CLASS with actual extensions of `Error` instead of throwing arrays to the wind.

(I still need to test this in CODE before merging)

### DIFF
--- a/public/components/ajax-step-flow/ajax-step-flow__slide.js
+++ b/public/components/ajax-step-flow/ajax-step-flow__slide.js
@@ -1,8 +1,8 @@
 // @flow
 
-import Raven from 'raven-js';
-import { showErrorText } from '../form/form-feedback-wrap';
-import { getUrlErrors } from '../../js/get-url-errors';
+import { showErrorText } from 'components/form/form-feedback-wrap';
+import { getUrlErrors } from 'js/get-url-errors';
+import { getRaven } from 'components/sentry/sentry';
 import {
   formRoutes as validAjaxFormRoutes,
   linkRoutes as validAjaxLinkRoutes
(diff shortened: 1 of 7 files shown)


public/components/ajax-step-flow/ajax-step-flow__slide.js
```diff
@@ -15,14 +21,19 @@ const SLIDE_STATE_DEFAULT: string = 'SLIDE_STATE_DEFAULT';
 
 const EV_DONE: string = 'form-done';
 
-const ERR_MALFORMED_HTML: string = 'ERR_MALFORMED_HTML';
-const ERR_MALFORMED_RESPONSE: string = 'ERR_MALFORMED_RESPONSE';
-const ERR_BACKEND_ERROR: string = 'ERR_BACKEND_ERROR';
+type ParsedResponse = {
+  type: string,
```
walaura: this should be an enum

public/components/ajax-step-flow/ajax-step-flow__slide.js
```diff
@@ -80,42 +91,54 @@ const fetchSlide = (
...
+          url: response.url
+        };
+      }
+    })
+    .then((parsedResponse: ParsedResponse) => {
+      if (parsedResponse.type === 'json' && parsedResponse.body.returnUrl) {
+        window.location.href = parsedResponse.body.returnUrl;
+        return new Promise(() => {});
+      } else if (parsedResponse.body.html) {
+        return [parsedResponse.body.html, parsedResponse.url];
```
walaura: might as well return `ParsedResponse ` now that we have that

public/components/ajax-step-flow/ajax-step-flow__slide.js
```diff
@@ -152,13 +142,20 @@ const fetchAndDispatchSlide = (
...
-      dispatchDone($slide, {
-        $slide: getSlideFromFetch(responseHtml),
-        url,
-        reverse: props.reverse
-      })
-    )
+    .then(parsedResponse => {
+      if (parsedResponse.type === 'json' && parsedResponse.body.returnUrl) {
+        window.location.href = parsedResponse.body.returnUrl;
+        return new Promise(() => {});
```
rahilb: ???

PR #445, opened 2018-07-25
cf deploy and temporary http origin

QuarpT:
- cf riffraff deploy
- temporarily enable http for identity frontend while changing certificates

### DIFF
--- a/riff-raff.yaml
+++ b/riff-raff.yaml
@@ -21,3 +21,14 @@ deployments:
           Recipe: identity-base
           AmigoStage: PROD
       amiEncrypted: true
+  cfn:
+    type: cloud-formation
+    parameters:
+      cloudFormationStackName: IdentityFrontend
+      templatePath: identity-frontend.yaml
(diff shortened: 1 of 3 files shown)


riff-raff.yaml
```diff
@@ -10,6 +10,22 @@ deployments:
...
+      - cfn
+
+  cfn:
+    type: cloud-formation
+    parameters:
+      cloudFormationStackName: IdentityFrontend
+      templatePath: identity-frontend.yaml
+      prependStackToCloudFormationStackName: false
+      cloudFormationStackByTags: false
+      amiParametersToTags:
```
jfsoul: is this needed?

riff-raff.yaml
```diff
@@ -10,6 +10,22 @@ deployments:
       prefixStack: false
     dependencies:
       - update-ami
+      - cfn
+
+  cfn:
```
jfsoul: slightly unnecessary shortname?

PR #451, opened 2018-08-01
Error message toast

rahilb:
Make a toast box for reporting errors consistently, see [the design](https://www.figma.com/file/FnjR7V5qgM9eto05QPoB8gNY/The-Guardian-component-library?node-id=882%3A4) for more details. 

screen shot 2018-08-01 at 13 39 10

No JavaScript? No problem! You get the following: 

screen shot 2018-08-01 at 13 44 14


rahilb:
It looks a bit naff on mobile for reasons: I'm fixing it now... 

screen shot 2018-08-01 at 14 03 21

rahilb:
>It looks a bit naff on mobile

fixed in aa1408c

screen shot 2018-08-01 at 14 11 13

walaura:
👏👏👏👏  happy to have a look at the css later today

public/css/globals/_mixins.css
```diff
@@ -24,16 +24,16 @@
   }
 }
 
-@define-mixin link {
```
walaura: Can we set defaults for this? `@define-mixin link $link-color: var(--color-link), $decoration-color:var(--color-link-underline)` _might_ work [according to this](https://github.com/postcss/postcss-mixins) but i'm actually quite unsure due to the way css variables are set up

public/css/globals/_vars.css
```diff
@@ -26,13 +26,13 @@
...
   --color-cta: #ffe500;
-
+  --color-white: #fff;
   --color-border: #abc2c9;
   --color-border-hover: #839ca3;
-
   --color-text: #121212;
 
   --color-error: #c70000;
+  --color-error-rgba: rgba(199, 0, 0, 0.8);
```
walaura: instead of setting a new variable we can use `color-mod` with `--color-error` like this
```
color: color-mod(var(--color-error) alpha(80%));
```

public/components/error-toast/_error-toast.global.css
```diff
@@ -0,0 +1,77 @@
+@import 'css/_globals.css';
+
+.error-toast {
+  display: flex;
+  background-color: var(--color-error-rgba);
+  color: white;
+  overflow: hidden;
+  @mixin font-button;
```
walaura: this should probably be `font-body-copy`

public/components/error-toast/_error-toast.global.css
```diff
@@ -0,0 +1,77 @@
...
+    & .error-toast__msg__report {
+      @mixin link var(--color-white) var(--color-link-white);
+      display: none;
+    }
+  }
+
+  & .error-toast__icon {
+    flex: 0 0 auto;
+    display: inline-block;
+    width: 15%;
```
walaura: this width should be absolute, ideally a multiplier of `baseline-gutter` for tidyness

public/components/error-toast/_error-toast.global.css
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    @media (--viewport-min-tablet) {
+      bottom: calc(var(--size-baseline-vertical) * 4);
+      margin-left: calc(var(--size-baseline-gutter) * 6);
+    }
+
+    @media (--viewport-min-desktop) {
+      margin-left: calc(var(--size-baseline-gutter) * 4);
+    }
```
walaura: I'm not a fan of all this viewport juggling. can we use `vw` and `vh` for `bottom` and `left` instead? that should give us a very similar effect

public/components/form/form-feedback-wrap.js
```diff
@@ -3,7 +3,7 @@ import { localisedError } from 'js/config';
 
 const $elements: HTMLElement[] = [];
 
-const selector: string = '.form-feedback-wrap';
```
walaura: we should rename this file accordingly

public/layout.hbs
```diff
@@ -51,7 +58,7 @@
   </main>
 
   {{> components/footer/_footer }}
-
+  <div class="error-toast-ajax-wrap" />
```
walaura: I think this has to be `<div class="error-toast-ajax-wrap"></div>`. As far as I know, divs can't be self-closing. Even if they can be this style could be confusing for whoever comes next

QuarpT:
looks good

public/components/error-toast/_error-toast.global.css
rahilb: I don't know what that means

public/css/globals/_mixins.css
rahilb: I did it and nothing seems broken so I think  it does work.

public/components/error-toast/_error-toast.global.css
walaura: if you use 
```
left: 5vw;
bottom: 5vw;
```

instead of multiples of gutter

the toast will always be at 5% away from the viewport

walaura:
😍

PR #463, opened 2018-08-23
Add data-link-name for jobs checkbox tracking

laurenemms:
Added functionality to recently added jobs consent checkbox on registration page to enable tracking via Ophan and GA

### DIFF
--- a/public/components/register-form/_form-field-consent.hbs
+++ b/public/components/register-form/_form-field-consent.hbs
@@ -1,5 +1,5 @@
 {{#each consents as |consent|}}
-<div class="form-checkbox form-field-wrap">
+<div class="form-checkbox form-field-wrap" data-link-name="consent : {{consent.id}}">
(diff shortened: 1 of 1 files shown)


public/components/register-form/_form-field-consent.hbs
```diff
@@ -1,5 +1,5 @@
 {{#each consents as |consent|}}
-<div class="form-checkbox form-field-wrap">
+<div class="form-checkbox form-field-wrap" data-link-name="consent : {{consent.id}}">
```
rahilb: I think this will send an event every time the checkbox is clicked, regardless of the previous value... ideally we'd only send it when it transitions from unticked to ticked... I think to do that you'd have to change the file `analytics-in-page-click.js`.

jfsoul: Yeah, we decided to first cater for the simple case by looking at GA and isolating the event flow of `single consent click -> successful registration`.  We'd ignore the ones where there was more than one click of the box.

PR #468, opened 2018-09-04
Disable password manager one step signin, hopefully fix lastpass

rahilb:
tested locally: lastpass can fill the email (on the first step) and the password (on the second step).


rahilb:
tested with lastpass in CODE, same results as local (works as intended)

PR #471, opened 2018-09-05
Update CODE ssl certificate

rahilb:


QuarpT:
why do we need a different certificate for CODE?

PR #473, opened 2018-09-05
Enables fatal warnings. 

rahilb:
Does not enable the nice flags that actually catch bugs (yet), as when I did there were many errors that I will fix in later PRs.


test/com/gu/identity/service/client/ApiRequestSpec.scala
```diff
@@ -30,12 +30,12 @@ class ApiRequestSpec extends PlaySpec with MockitoSugar{
 
     "return a iterable containing the X-GU-ID-Client-Access-Token and value for this token" in {
       val headers = Seq(ApiRequest.apiKeyHeader)
-      headers.head equals ("X-GU-ID-Client-Access-Token","Bearer ##key##")
+      headers.head mustEqual "X-GU-ID-Client-Access-Token" -> "Bearer ##key##"
```
rahilb: this was particularly embarrassing: these tests could never fail.

QuarpT: shocking

rahilb:
> return something different to the users for ClientUnauthorizedError 

That would be great, but it seems to do that is a ton of useless boilerplate, so I'd rather just wait for the refactoring of all of this.

PR #475, opened 2018-09-05
Warn for unused params, remove all unused params

rahilb:


app/com/gu/identity/frontend/models/UrlBuilder.scala
```diff
@@ -38,7 +38,7 @@ object UrlBuilder {
     apply(baseUrl, buildParams(Some(returnUrl), skipConfirmation, clientId, group, skipThirdPartyLandingPage))
 
   def apply(baseUrl: String, returnUrl: ReturnUrl, skipConfirmation: Option[Boolean], clientId: Option[ClientID], group: Option[String], skipThirdPartyLandingPage: Option[Boolean], skipValidationReturn: Option[Boolean]): String =
-    apply(baseUrl, buildParams(Some(returnUrl), skipConfirmation, clientId, group, None, skipValidationReturn))
```
rahilb: This was another embarrassing one: the param was provided but not passed to the function 🙈

rahilb:
we can't enable warnings for unused-privates, as the java handlebars wrapper is using them, via runtime reflection (which of course the compiler is unable to pick up).

PR #478, opened 2018-09-12
Fix sign in on iOS8

walaura:
we had a bit of polyfill troubles. it's all good now.

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,1 +1,3 @@
 IE >= 9
+iOS >= 8
+not dead
--- a/package-lock.json
+++ b/package-lock.json
@@ -2501,6 +2501,11 @@
         }
       }
     },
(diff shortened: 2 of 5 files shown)


.browserslistrc
```diff
@@ -1 +1,3 @@
 IE >= 9
+iOS >= 8
```
rahilb: Does this mean iOS versions 8 or more will be polyfilled?

walaura: yep

package-lock.json
```diff
@@ -2501,6 +2501,11 @@
         }
       }
     },
+    "css-unit-converter": {
```
rahilb: is this actually used in this PR ?

walaura: package-lock is a bit of a mess it should have been commited before

PR #480, opened 2018-09-21
Added INTCMP parameter for tracking from redirects to /signin

laurenemms:
Added the functionality to ensure that the redirection from /agree/:group to /signin and /signin/:type to /signin hold onto the INTCMP parameter.


app/com/gu/identity/frontend/models/UrlBuilder.scala
```diff
@@ -60,6 +60,9 @@ object UrlBuilder {
 
   def apply(baseUrl: String, call: Call): String = s"$baseUrl${call.url}"
 
+  def apply(baseUrl: String, intcmp: Option[String]): String =
```
jfsoul: I would probably change the `baseUrl: String` to a `call: Call` (and do the `call.url` in this method), but that's super picky

jfsoul:
Don't forget to rename this PR too 😄

PR #484, opened 2018-10-29
Replace GU_SIGNIN_EMAIL with value in Session Flash

Mullefa:
Since we are only using the cookie in the next request, this seems like a perfect use case for [ScalaSessionFlash](https://www.playframework.com/documentation/2.6.x/ScalaSessionFlash). This solves the problem of the `GU_SIGNIN_EMAIL` cookie lasting for a whole session, which potentially lead to confusing behaviour for the user.

I also took the time to refactor the action `emailSignInFirstStep` - motivation included in the comments.

### DIFF
--- a/app/com/gu/identity/frontend/authentication/CookieService.scala
+++ b/app/com/gu/identity/frontend/authentication/CookieService.scala
@@ -58,18 +58,4 @@ object CookieService {
 
     }
   }
-
-  def signInEmailCookies(email: String)(config: Configuration): Seq[PlayCookie] = {
-    Seq(PlayCookie(
-      name = "GU_SIGNIN_EMAIL",
-      value = email,
-      maxAge = None,
-      path = "/",
-      domain = Some(config.identityCookieDomain),
-      secure = true,
-      httpOnly = false
-    )
-    )
-  }
-
 }
--- a/app/com/gu/identity/frontend/controllers/SigninAction.scala
+++ b/app/com/gu/identity/frontend/controllers/SigninAction.scala
@@ -111,17 +111,15 @@ class SigninAction(
       case Left(errors) => Left(errors)
       case Right(response) =>
         signInFirstStepMetricsLogger(request)
-        val emailLoginCookie = CookieService.signInEmailCookies(body.email)(config)
         Right(
           successfulFirstStepResponse(
             response.userType,
             successfulReturnUrl,
-            emailLoginCookie,
             body.skipConfirmation,
             body.clientId,
             body.groupCode,
             body.skipValidationReturn
-          )
+          ).flashing("email" -> body.email) // used by endpoint /signin/:signInType / Application.twoStepSignInChoices
         )
     }
   }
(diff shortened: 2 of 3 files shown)


app/com/gu/identity/frontend/authentication/CookieService.scala
```diff
@@ -58,18 +58,4 @@ object CookieService {
 
     }
   }
-
-  def signInEmailCookies(email: String)(config: Configuration): Seq[PlayCookie] = {
```
Mullefa: no longer needed since Session Flash is now being used instead of a cookie

app/com/gu/identity/frontend/controllers/SigninAction.scala
```diff
@@ -97,29 +97,30 @@ class SigninAction(
     signInAction((_, cookies) => successfulSmartLockSignInResponse(cookies), successfulAjaxSignInResponse, signInSmartLockMetricsLogger)
   }
 
-  def emailSignInFirstStep = SignInServiceAction(bodyParser) { req =>
-    emailSignInFirstStepAction(successfulFirstStepResponse, signInFirstStepMetricsLogger)(req)
-  }
+  def emailSignInFirstStep: Action[SignInActionRequestBody] = SignInServiceAction(bodyParser) { request =>
 
-  def emailSignInFirstStepAction(successResponse: (String, ReturnUrl, Seq[Cookie], Option[Boolean], Option[ClientID], Option[GroupCode], Option[Boolean]) => Result, metricsLogger: (Request[SignInActionRequestBody]) => Unit) = { implicit request: Request[SignInActionRequestBody] =>
```
Mullefa: Since this is only ever being called with one pair of functions - `successfulFirstStepResponse()` and `signInFirstStepMetricsLogger()` - remove this method and use the functions directly in place. A lot clearer IMO.

app/com/gu/identity/frontend/controllers/SigninAction.scala
```diff
@@ -97,29 +97,30 @@ class SigninAction(
...
     val body = request.body
 
-    lazy val returnUrl = body.returnUrl.getOrElse(ReturnUrl.defaultForClient(config, body.clientId))
+    val returnUrl = body.returnUrl.getOrElse(ReturnUrl.defaultForClient(config, body.clientId))
 
-    val successfulReturnUrl = body.groupCode match {
-      case Some(groupCode) =>
-        UrlBuilder.buildThirdPartyReturnUrl(returnUrl, body.skipConfirmation, skipThirdPartyLandingPage = true, body.clientId, groupCode, config)
-      case _ => returnUrl
+    val successfulReturnUrl = body.groupCode.fold(returnUrl) { groupCode =>
```
Mullefa: More idiomatic (and clearer 🤔) to fold on an option rather than pattern match.

app/com/gu/identity/frontend/controllers/SigninAction.scala
```diff
@@ -97,29 +97,30 @@ class SigninAction(
...
+      UrlBuilder.buildThirdPartyReturnUrl(returnUrl, body.skipConfirmation, skipThirdPartyLandingPage = true, body.clientId, groupCode, config)
     }
 
     identityService.getUserType(body).map {
```
Mullefa: with a view to _maybe_ introducing cats to this project in the future (😉), if this was an `EitherT` then [map](signInFirstStepMetricsLogger) would be simpler i.e.
```
identityService.getUserType(body).map(response => successfulFirstStepResponse(...)
```

Mullefa:
Build failing for reasons I'm not too clear about:
```
[18:44:22][Step 4/5] java.lang.RuntimeException: problem during retrieve of org.scala-sbt#boot-app: java.io.FileNotFoundException: /opt/teamcity/.sbt/boot/scala-2.12.4/org.scala-sbt/sbt/1.1.2/util-control_2.12-1.1.3.jar (No such file or directory)
[18:44:22][Step 4/5]   (see /opt/teamcity/.sbt/boot/update.log for complete log)
```

app/com/gu/identity/frontend/controllers/SigninAction.scala
Calum-Campbell: Probably more idiomatic, wether it's clearer I think depends on your FP/scala experience. I like it though 👍

PR #485, opened 2018-10-30
update unsubscribe confirmation wording to account for eventual consistency

QuarpT:
update unsubscribe confirmation wording to account for eventual consistency

Consent unsubscribes take time to propagate between Identity > Data Lake > Braze


conf/messages.en
```diff
@@ -148,7 +148,7 @@ changeEmailSuccessful.unexpectedTitle=Sorry there was a problem
 
 email.unsubscribe.pageTitle=Unsubscribe Confirmation
 email.unsubscribe.title=Unsubscribe Confirmation
-email.unsubscribe.description=You have been unsubscribed from this email.
+email.unsubscribe.description=You will be unsubscribed from this email within the next 24 hours.
```
jfsoul: I am not sure which is better but here is an alternative:

`You have been unsubscribed from this email.  The change may take up to 24 hours to take effect.`

PR #492, opened 2018-11-20
Setup for S&C sign-in page

laurenemms:
When designs come through, CSS can go in _simple-and-coherent.global.css. Variations to sign in page use layout specified in two-step-signin-start-page.hbs and text is mapped in Text.scala.

### DIFF
--- a/app/com/gu/identity/frontend/models/ClientID.scala
+++ b/app/com/gu/identity/frontend/models/ClientID.scala
@@ -46,8 +46,14 @@ case object GuardianCommentersClientID extends ClientID {
   val id = "comments"
 }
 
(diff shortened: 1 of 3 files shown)


app/com/gu/identity/frontend/models/ClientID.scala
```diff
@@ -46,8 +46,14 @@ case object GuardianCommentersClientID extends ClientID {
   val id = "comments"
 }
 
+case object SimpleAndCoherentClientID extends ClientID {
+  val id = "simpleAndCoherent"
+
+  override def hasSkin = true
```
Mullefa: can be a `val`

PR #493, opened 2018-11-20
Remove signin email on login and logout

Calum-Campbell:
- Remove GU_SIGNIN_EMAIL cookie after successful login
- Remove GU_SIGNIN_EMAIL on successful logout
- Also start to remove SC_GU_LA on logout (should have been doing this anyway)
NB:
Decided to leave the signin_email as a session cookie, this way it will still be around if the user navigates away from profile then back to it later. I don't think it is necessary to enforce it's deletion after a certain time.

### DIFF
--- a/test/com/gu/identity/frontend/controllers/SigninActionSpec.scala
+++ b/test/com/gu/identity/frontend/controllers/SigninActionSpec.scala
@@ -123,6 +123,9 @@ class SigninActionSpec extends PlaySpec with MockitoSugar {
       status(result) mustEqual SEE_OTHER
       redirectLocation(result) mustEqual returnUrl
 
+      // successful sign in should set an unset cookie for GU_SIGNIN_EMAIL cookie
+      resultCookies.count(cookie => cookie.name == "GU_SIGNIN_EMAIL" && cookie.maxAge.contains(0)) mustEqual 1
+
       resultCookies.size mustEqual 2
       resultCookies.head mustEqual testCookie
     }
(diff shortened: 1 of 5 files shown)


test/com/gu/identity/frontend/controllers/SigninActionSpec.scala
```diff
@@ -123,7 +123,7 @@ class SigninActionSpec extends PlaySpec with MockitoSugar {
       status(result) mustEqual SEE_OTHER
       redirectLocation(result) mustEqual returnUrl
 
-      resultCookies.size mustEqual 1
+      resultCookies.size mustEqual 2
```
Mullefa: Appreciate the tests are pre-existing, but maybe worth testing that we are setting an unset cookie explicitly? Not sure what a count of 2 gives us?

Calum-Campbell:
@Mullefa _"How is the cookie getting unset on user sign out?"_
By adding the cookie names to the "known cookies" in app/com/gu/identity/frontend/authentication/AuthenticationService.scala
These are all removed here https://github.com/guardian/identity-frontend/blob/master/app/com/gu/identity/frontend/authentication/AuthenticationService.scala#L38

test/com/gu/identity/frontend/controllers/SigninActionSpec.scala
Calum-Campbell: Yep this is a good shout, will add something now

Mullefa:
Got ya. Thanks for the explainer.

test/com/gu/identity/frontend/controllers/SigninActionSpec.scala
Mullefa: Nice!

Calum-Campbell:
@Mullefa _"How is the cookie getting unset on user sign out?"_
By adding the cookie names to the "known cookies" in app/com/gu/identity/frontend/authentication/AuthenticationService.scala
These are all removed

PR #495, opened 2018-11-22
Separate consent unsubscribe email

QuarpT:
The consent unsubscribe landing page requires wording depending on the consent.


QuarpT:
picture 86

build.sbt
```diff
@@ -24,6 +24,7 @@ libraryDependencies ++= Seq(
   ws,
   filters,
   "org.scalatestplus.play"          %%  "scalatestplus-play"        %   "3.1.2"   %   Test,
+  "org.mockito"                     % "mockito-all"                 % "1.9.5"     %   Test,
```
QuarpT: without this the build fails 
I'm a bit confused why this doesn't get pulled in by scalatest 😕

PR #497, opened 2018-12-11
New sign in page

laurenemms:
After a request from S&C to implement a sign in skin for their payment failure journey, we took this as an opportunity to redesign our sign in pages altogether. However, changes to the sign in layout affects the layout of all pages within the Identity-Frontend project. 

This PR contains changes to the new sign in pages including: 
- Sign in component aligned to the left 
- Billionaire SVG on the right side of the screen for simple and coherent
- Some wording changes
- Facebook icon changes and order of social links
- Formatting alterations (wrapping each content in layout-section div)
- Updates to fonts and colours
- Sign in component internal changes (including green checkbox, yellow header, back and next buttons, boxed input fields, etc)
- Removal of animations 
- Checks if an SVG is attached to a clientId and if it is, passes it to the layout. So far, only the clientId=subscriptions has an SVG
- If no SVG is present, centralises the sign in component 

Pages using new layout
- /signin
- /signin/new
- /signin/guest
- /signin/current
- /reset & /reset/resend
- /agree

As we only have designs for the sign in flow, this PR also creates two new layout files (one for hbs and one for css) to allow for pages without designs to continue using the old layout until designs are made. 

When we receive new designs and they are ready to be implemented
- Go through files using old layout and change partial from layout-old to layout
- Remove --oldLayout from class names in pages that use old layout
- Delete _layout-old.global.css
- Remove _layout-old.global.css import from load-global-css.js
- Delete layout-old.hbs


app/com/gu/identity/frontend/views/models/LayoutViewModel.scala
```diff
@@ -47,6 +47,11 @@ case class SkinInViewModel(
   banner: Boolean = false
 )
 
+case class SvgViewModel(
```
walaura: could this be `name: Option[String]` instead of having a separate `exists` boolean?

public/components/form/_form-button.global.css
```diff
@@ -111,24 +117,29 @@
     color: var(--color-text);
     &:hover,
     &:focus {
+      background-color: #f6f6f6;
```
walaura: can this be a colour variable instead?>

public/components/form/_form.global.css
```diff
@@ -84,24 +98,32 @@ generic input wrapper
 }
 
 .form-field-wrap > label {
+  :not(.form-field--email)
```
walaura: i don't think this is valid css syntax?

public/components/layout/_layout-old.global.css
```diff
@@ -0,0 +1,69 @@
...
+& .form-button {
+@media (--viewport-min-tablet) {
+  display: inline-flex;
+  vertical-align: 10%;
+  margin-left: calc(var(--size-baseline-gutter) / 2);
+}
+}
+&:not(.layout-header__title--standfirst) {
+@media (--viewport-max-mobile-medium) {
+  padding-right: 3em; /*avoid overlaps with the logo*/
```
walaura: out of scope for this PR but I don't think we need this anymore as the logo is no longer popping out of the header

public/components/layout/_layout.global.css
```diff
@@ -27,8 +115,11 @@
 }
 
 .layout-header {
-  padding: var(--size-baseline) 0 var(--size-baseline-vertical);
+  padding: 0;
+  background-color: var(--color-cta);
```
walaura: I'd make a separate `--color-header` variable even if it is the same colour just so if we were to tweak the header colour it doesn't affect the cta's

public/components/two-step-signin/two-step-signin-new.hbs
```diff
@@ -1,37 +1,33 @@
...
+      <label class="form-field-wrap__title form-field--email" for="tsse-email">{{twoStepSignInPageText.emailTitle}}</label>
+      <a class="form-field-wrap--email-change" data-is-reverse="true" href="{{signinUrl}}">{{twoStepSignInPageText.changeEmailLinkShort}}</a>
       <input class="form-input form-field-wrap__field" type="email" name="email" id="tsse-email" autocomplete="username" autocapitalize="off" autocorrect="off" spellcheck="false" readonly value="{{email}}" />
     </div>
 
     {{> components/register-form/_form-field-name text=twoStepSignInPageText.registerForm }}
 
     {{> components/register-form/_form-field-phone
-      text=twoStepSignInPageText.registerForm askForPhoneNumber=askForPhoneNumber countryCodes=countryCodes
+      text=twoStepSignInPageText.registerForm askForPhoneNuxmber=askForPhoneNumber countryCodes=countryCodes
```
walaura: typo!

public/email-unsubscribe-page.hbs
```diff
@@ -2,21 +2,21 @@
 
 {{# partial "content" }}
 
-  <section class="layout-wrap-width layout-wrap-width--wide">
+  <section class="layout-wrap-width--oldLayout layout-wrap-width--wide--oldLayout">
```
walaura: the way BEM is meant to work, this class name would have to be both `class="layout-wrap-width layout-wrap-width--oldLayout"`. Then `oldLayout` would override whatever it needs. I don't think this is that important here as we are eventually get rid of these altogether but something worth keeping in mind for the future!

app/com/gu/identity/frontend/views/models/LayoutViewModel.scala
laurenemms: Ah - I did try to do that at first but couldn't find a way depend conditions on Options in hbs, I could only see a way to depend it on a boolean value. Is there a way to do this in handlebars?

public/components/form/_form.global.css
laurenemms: Weirdly it works but if there is a better way of doing it then let me know. Basically we want all labels of form-field-wrap to be display: block except for that with a class of .form-field--email. The reason behind this is because we want the change link to appear next to it rather than under.

**without :not(.form-field--email)**
screen shot 2019-01-14 at 15 04 57

**with :not(.form-field--email)**
screen shot 2019-01-14 at 15 04 46

public/components/layout/_layout-old.global.css
laurenemms: Just checked and you're right, both 'layout-header__title--standfirst' and 'layout-header__title--has-proxy' aren't used anymore and can both be removed. I'll do in a separate PR

public/components/layout/_layout.global.css
laurenemms: Will do - thanks!

public/email-unsubscribe-page.hbs
laurenemms: Oh great, thanks for telling me. Yeah this will hopefully be deleted soon but I will make sure to do this next time.

laurenemms:
> sprinkled a couple comments in for now, will try to run it locally later today!

Thanks very much! Have put some follow up questions and let me know if you run into any issues running it later. Thank you!

app/com/gu/identity/frontend/views/models/LayoutViewModel.scala
walaura: ah! i think it will be like javascript where its a `string` or `null` so `if (name) { name }` (but in hbs syntax) will work? But I'm not sure to be honest

public/components/form/_form.global.css
walaura: ah, that implementation is perf then, (maybe could be abstracted and called something like `form-field--with-link`? but not a biggie) I'm gonna have a look at to _how_ it's working  because I'm intrigued

app/com/gu/identity/frontend/views/models/LayoutViewModel.scala
laurenemms: It worked! Thanks!

public/components/form/_form.global.css
laurenemms: Okay - thanks for all your comments!

public/components/layout/_layout.global.css
```diff
@@ -125,9 +211,16 @@ texty text
...
   }
   &.layout-text--bold {
-    font-weight: 500;
+    font-weight: 600;
+    padding: 10px 0;
   }
   & + .layout-text {
     margin-top: calc(var(--size-baseline-vertical) / 2);
   }
+  &.layout-text--new {
```
walaura: can we unify this with the normal layout text or give it a more descriptive name?

public/components/form/_form-button.global.css
```diff
@@ -111,24 +117,29 @@
...
+      background-color: var(--color-button-hover);
     }
   }
   &.form-button--oauth-google {
-    color: #fff;
-    background-color: #4285f4;
+    color: var(--color-text);
+    font-weight: 600;
+    border: 1px solid var(--color-border);
+    margin: var(--size-baseline-vertical) 0;
```
walaura: I'm not super happy about hardcoding this margin here, can we put this in the element that contains both OAuth buttons? Personally I like doing that with all size/margin/position properties so the components can be placed anywhere on the page without bringing layout into it

public/components/form/_form.global.css
walaura: Can we turn this into `.form-field-wrap > label:not(.form-field--email)` I have no idea how the current syntax is working but i think it's a bug in postcss?

walaura:
I've got MORE! That said, please use your criteria as to when to just merge this and address the rest later (or not)

PR #498, opened 2019-01-07
New sign-in endpoint with email auto-filled

Mullefa:
__TODO:__ before merging, update the identity API token used by identity frontend with the scope `signinTokens.r`

Work is in progress to send payment failure emails via identity so that an encrypted email token can be included in the link within the email (see e.g. [#160](https://github.com/guardian/membership-workflow/pull/160) in membership-workflow and the [payment-failure lambda](https://github.com/guardian/identity-processes/tree/master/payment-failure) in identity-processes).

This PR provides a new endpoint - `/actions/signin/with-email-autofill` - where non-signed-in users with this token can re-directed to. This endpoint by-passes the first sign-in step and goes directly to the second step (with email autofilled).

PR #499, opened 2019-01-08
Consider social accounts with in email autofill endpoint

Mullefa:
__TODO__: 
- [x] merge [#1499](https://github.com/guardian/identity/pull/1499) in identity before merging this PR
- [x] update the identity API token used by identity frontend with the scope `users.socialLinks.r`

The work in #498 would redirect a user to the second sign-in step, even if they had a social account. This PR redirects users with a social account to the first sign-in step (so that they can sign-in with the Google or Facebook buttons, rather than a password - we presume this is what they want).


app/com/gu/identity/frontend/controllers/SigninAction.scala
```diff
@@ -166,6 +166,19 @@ class SigninAction(
...
 
+  def getResultForEmailAutofillSignIn(email: String, userSignInDetails: UserSignInDetails): Result =
+    if (userSignInDetails.hasGoogleSocialLink || userSignInDetails.hasFacebookSocialLink) {
+      // Presume that users with a social account would
+      // rather sign-in with the Google or Facebook buttons than a password.
+      // Therefore, send them to the first sign-in step.
+      SeeOther(redirectRoute)
+    } else {
+      // Otherwise send them to the second sign-in step (where email address will be auto-filled).
+      val userType = if (!userSignInDetails.hasAccount) NewUser else if (!userSignInDetails.hasPassword) GuestUser else CurrentUser
```
laurenemms: I was just wondering if you can use the getUserType in IdentityService here rather than have a conditional statement?

app/com/gu/identity/frontend/models/UserSignInDetails.scala
```diff
@@ -0,0 +1,14 @@
+package com.gu.identity.frontend.models
+
+import play.api.libs.json.{Json, Reads}
+
+case class UserSignInDetails(
+    hasAccount: Boolean,
+    hasPassword: Boolean,
```
laurenemms: Continuing from my last comment - do we need hasAccount and hasPassword as well as the getUserType method in IdentityService?

app/com/gu/identity/frontend/controllers/SigninAction.scala
Mullefa: `getUserType` doesn't provide sufficient information e.g. users with a social account and no password, and users with just an email address are both assigned user type `guest`, but in identity frontend we need to distinguish between these two types of users.

app/com/gu/identity/frontend/models/UserSignInDetails.scala
Mullefa: Given that we need the additional information provided by the user sign-in details endpoint (see previous comment), I would advocate for the deprecation of the user type endpoint. This would mean we have one canonical endpoint for information related to user sign-in, therefore (hopefully) reducing the confusion.

app/com/gu/identity/frontend/controllers/SigninAction.scala
laurenemms: Got it, thanks

app/com/gu/identity/frontend/models/UserSignInDetails.scala
laurenemms: Yes I agree, I'll add a card to the backlog on Trello for this

Mullefa: Thanks 👌

PR #505, opened 2019-01-16
Add console appender

Mullefa:
Facilitates work locally. Personally I find it a lot more convenient than writing logs to a file, but @Calum-Campbell have said this might not be the majority preference in identity (?) Happy to close the PR if this is the case.

Note: I'm using an async appender so performance is not impacted. 

As an additional bit of health work, I'd like to resolve errors messages of the following nature:

```
WARN  c.g.r.connection.AbstractConnection:84 - An exception due to the connection occurred, a lockdown will be initiated.
com.getsentry.raven.connection.ConnectionException: {"error":"Invalid project_id: u'70882'"}
...
```


jfsoul:
Up to you @Calum-Campbell and @laurenemms as the ones who work with it most often.

In my experience it depends how noisy the logs are.

Calum-Campbell:
My concerned were that the logs would be too noisy, I remember having this issue on an app before. Just tried this locally and it seems ok though. I am getting cloudwatch and sentry errors though, can these be avoided by updating the dev conf file at all?

Mullefa:
From developing on this app locally, I don't think noise is a concern. Especially if we resolve errors mentioned previously by me and Callum.

@laurenemms as a primary developer of this application, have you got any thoughts?

Mullefa:
From developing on this app locally, I don't think noise is a concern. Especially if we resolve errors mentioned previously by me and Callum.

@laurenemms as a primary developer of this application, have you got any thoughts?

laurenemms:
Does this mean that the logs will also be sent to the console as well as the identity-frontend-access.log file? If it does then I think I would find that useful. Sorry if I'm not understanding this correctly.

Mullefa:
Yeah basically. The application logs (i.e. the logs that are written to the `identity-frontend-app.log` file), would also be written to the console.

Mullefa:
[Bring the noise](https://www.youtube.com/watch?v=kl1hgXfX5-U)

PR #506, opened 2019-01-16
Drop encrypted email if not valid

Calum-Campbell:
If the encrypted email param fails, we wont to drop it from the query params from the next page view


Mullefa:
Just realised, in the case where there is a valid encrypted email token and the user has a social account [this](https://github.com/guardian/identity-frontend/blob/20269fb91a4cf784d568107d62a4457afef4c8e3/app/com/gu/identity/frontend/controllers/Application.scala#L86) will be evaluated. In particular (because they are not redirected) they will see the `encryptedEmail` parameter. 

Wondering if these re-directs are preferable to showing the query parameter? Not sure.

Calum-Campbell:
@Mullefa I think it makes sense to drop it where possible, so let's redirect here as well? Have updated.

PR #507, opened 2019-01-17
Only show social buttons for social users

laurenemms:
When a user receives an email with a link that takes them to sign in, we check if they have a social account and if they do then we prompt them to sign in via a social link (Facebook or Gmail) this is to lower the chance of users creating another account with an email address that is not linked to their payment details or other information.

Sign in for user with social account
screen shot 2019-01-17 at 16 37 35

After link to sign in with email is clicked 
screen shot 2019-01-17 at 16 37 28


app/com/gu/identity/frontend/views/ViewRenderer.scala
```diff
@@ -34,7 +34,8 @@ object ViewRenderer {
     clientId: Option[ClientID],
     group: Option[GroupCode],
     email: Option[String],
-    skipValidationReturn: Option[Boolean])
+    skipValidationReturn: Option[Boolean],
+    isSocialUser: Boolean = false)
```
Mullefa: do we need a default value here?

app/com/gu/identity/frontend/views/models/TwoStepSignInStartViewModel.scala
```diff
@@ -26,6 +26,7 @@ case class TwoStepSignInStartViewModel private(
 
   email:Option[String],
   skipValidationReturn: Boolean = false,
+  isSocialUser: Boolean = false,
```
Mullefa: do we need a default value here?

public/components/two-step-signin/two-step-signin-start.js
```diff
@@ -0,0 +1,15 @@
+const exp = document.getElementsByClassName('two-step-signin__expandable-link');
+let i;
+
+for (i = 0; i < exp.length; i += 1) {
```
Mullefa: Higher-level javascript methods are used in this code base. [`forEach()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach) might be appropriate.

laurenemms:
> A few minor comments.
> 
> Also, I wonder whether the query parameter should be declarative in name e.g. `socialButtonsOnly`? The difference is minor, but means for example we don't get potentially incorrect defaults `isSocialUser=false` (the user could have a social link) and makes sense in more contexts e.g. say we wanted to get more users to sign-up through social links (`socialButtonsOnly=true` is clearer than `isSocialUser=true`).
> 
> What d'you think?

No that's a good point, I haven't taken into account that case when I had named it. Will rename

public/components/two-step-signin/two-step-signin-start.js
laurenemms: Thank you!

PR #511, opened 2019-02-15
Use auto sign token if present

Mullefa:
Updates the `/signin` endpoint to attempt to use the `autoSignInToken` query parameter if present in the url.


app/com/gu/identity/frontend/services/IdentityService.scala
```diff
@@ -48,8 +49,9 @@ trait IdentityService {
   // It's possible the codebase would benefit from all methods within this service returning EitherT[Future, ServiceExceptions, A]
 
   def getUserType(email: String)(implicit ec: ExecutionContext): EitherT[Future, ServiceExceptions, UserType]
-  def decryptEmail(encryptedEmail: String)(implicit ec: ExecutionContext): EitherT[Future, ServiceExceptions, String]
-  def getUserSignInDetails(email: String)(implicit ec: ExecutionContext): EitherT[Future, ServiceExceptions, UserSignInDetails]
+  def decryptEmail(encryptedEmail: String)(implicit ec: ExecutionContext): EitherT[Future, AppException, String]
```
Mullefa: The `IdentityClient` returns a sequence of errors which is why the errors values in this class are typically `ServiceExceptions` (type alias for `Seq[AppException]`). However, a single `AppException` is a lot more convenient so the pre-existing:
```scala
SeqAppExceptions(errors: Seq[AppException]) extends AppException
```
is used to flatten the app exceptions.

app/com/gu/identity/frontend/services/IdentityService.scala
```diff
@@ -226,43 +228,64 @@ class IdentityServiceImpl(config: Configuration, adapter: IdentityServiceRequest
...
+          // when converting IdentityClientError(s) to AppException(s).
+          logger.error(s"unable to use auto sign-in token - errors: $errors")
+          appExceptionFromIdentityClientErrors(errors)
+        },
+        response => {
+          val cookies = response.cookies
+          val identityApiCookies = cookies.values.map(IdentityApiCookie(_, cookies.expiresAt))
+          // Setting rememberMe to false ensures the cookies returned will be session cookies.
+          // We don't want to make cookies that were created from an auto sign-in link permanent.
+          // TODO: why?
```
laurenemms: Do we know why now? Can we remove this?

PR #513, opened 2019-02-22
Include ab test parameters on redirects

Mullefa:
Adds the query parameters `abName` and `abVariant` to the `SignInRequestData` class so that they can be added to the url on redirects. This means we can use the query parameters field on Ophan to calculate sign-in conversion rates of the encrypted email test. See [#22](https://github.com/guardian/identity-processes/pull/22) in identity-processes for more context.

As alluded to in a code comment, we could do something more heavy weight in e.g. fire Ophan events client side to record (1) a user landing on a sign-in page when in the test; and (2) signing in successfully when in the test (which is more idiomatic); however for now I have gone for expediency (since it is not clear how many more of these tests will be run in the future).


app/com/gu/identity/frontend/models/UrlBuilder.scala
```diff
@@ -63,6 +63,9 @@ object UrlBuilder {
   def apply(call: Call, intcmp: Option[String]): String =
     apply(call.url, buildParams(intcmp = intcmp))
 
+  def apply(call: Call, intcmp: Option[String], abTest: Option[AbTest]): String =
+    apply(call.url, buildParams(intcmp = intcmp, abTest = abTest))
+
```
laurenemms: I wonder if we could build on the apply method on line 14 or make another that handles a call and _optional_ UrlParameters. I think I added the method above this one on line 63 because the `def apply(call: Call, params: UrlParameters): String =
    apply(call.url, params)` one didn't deal with optional params, and therefore couldn't have worked with INTCMP. Now we have two cases of optional params do you think it would be worth doing?

PR #519, opened 2019-03-04
Stop indexing of sign in page variants

jfsoul:
We're getting hit particularly hard recently by crawlers (especially apple bot) because they're following links on comments to `/signin?returnUrl=https://discussion.theguardian.com/comment-permalink/{id}`.  This should prevent that, but may also affect SEO of other variants of /signin (like the link from the header of dotcom).  Should this be more specific?


Mullefa:
If possible I think we should be more specific. [Here](https://stackoverflow.com/questions/9149782/ignore-urls-in-robot-txt-with-specific-parameters) are some examples of disallowing certain query strings, but I'm not sure if Apple supports the `*` matcher? i.e. it isn't mentioned [here](https://support.apple.com/en-gb/HT204683). Maybe try using it and see if it stops the traffic from apple bot?

PR #522, opened 2019-03-19
Consent registration AB Test

Calum-Campbell:
- Add supporter consent box to registration flow
- Add A/B test to display this to 50% of users with mvt_id cookie
- Add google analytics metric for components loaded with data-link-name attribute in order to track success of A/B test

image
image


app/com/gu/identity/frontend/mvt/MultiVariantTestID.scala
```diff
@@ -11,7 +11,7 @@ case class MultiVariantTestID(
 
 object MultiVariantTestID {
   val MVT_COOKIE_NAME = "GU_mvt_id"
-  val MAX_ID = 899999
+  val MAX_ID = 1000000
```
Calum-Campbell: This needs to be in line with frontends setting of the mvt_id as that is where this cookie is set. The max_id was bumped up in this PR https://github.com/guardian/frontend/pull/13269/files

walaura:
design nits: can we use a slightly less bold weight of text sans? This one is _super_ bold and we don't really use it across graun cc @zeftilldeath

Calum-Campbell:
@walaura thanks for the comment! Have run over it with @zeftilldeath , he's happy for it to stay as is for now but has given a couple of visual tweeks that I have applied.

public/components/analytics/analytics-component-load.js
```diff
@@ -0,0 +1,13 @@
+// @flow
+
+import { customMetric } from './ga';
+
+const selector: string = '*[data-link-name]';
+
+const init = ($component: HTMLElement): void => {
+  if ($component.dataset.linkName) {
+    customMetric({ name: $component.dataset.linkName, type: 'componentLoad' });
```
jfsoul: Do we get this for free with ophan or we could add it?

Calum-Campbell: Thanks yep have added ophan tracking for consents component rendering now.

zeftilldeath:
As far as I know we only use BOLD or REGULAR weights of text Sans Laura so it needs to be Bold, annoyingly the medium weight of the font is so close to regular at small sizes the difference wasn't obvious enough.

PR #524, opened 2019-03-25
Add payment failure text

laurenemms:
[https://trello.com/c/xyXsjlNT/1308-magic-links-create-clientid-for-payment-failure-and-add-new-message-to-the-sign-in-page](url)

If an auto sign in token is expired or invalid, the user is taken to the second step of sign in with their email pre-filled. This adds a new 'journey' parameter which, when set to 'paymentFailure', will display some text to the user reminding them why they are signing in. 

I was originally planning to only add the text to the second step of sign in, since the encrpytedEmail parameter will redirect users there and this parameter should be present if they have received the payment failure email. However, since we take users to first step of sign in if they have a social account, I have included it on both pages.

Social User
Screen Shot 2019-03-27 at 15 19 00

Not social user
Screen Shot 2019-03-27 at 15 19 20

With clientId set to subscriptions
Screen Shot 2019-03-27 at 15 23 35


app/com/gu/identity/frontend/controllers/Application.scala
```diff
@@ -39,6 +39,7 @@ class Application(
     socialOnlyButtons: Boolean,
     emailToken: Option[String],
     autoSignInToken: Option[String],
+    journey: Option[String],
```
laurenemms: I decided not to create a case class for Journey for now as it’s unclear as to whether we are going to use this parameter in the future for further adaptions to sign in, this can be reconsidered if we have more cases like this.

laurenemms:
> LGTM 👍.
> Don't know if it needs to go past design, maybe just let @zeftilldeath have a quick look over it.

Thanks! I have re-requested review as there are a number of changes after changing from using clientId

app/com/gu/identity/frontend/controllers/Application.scala
```diff
@@ -220,10 +226,11 @@ class Application(
       skipValidationReturn: Option[Boolean],
       encryptedEmail: Option[String],
       socialButtonsOnly: Boolean,
-      autoSignInToken: Option[String]
+      autoSignInToken: Option[String],
+      journey: Option[String]
```
jfsoul: Maybe put a comment in somewhere explaining the purpose of the journey parameter?

laurenemms:
> If you can get the opinion of a designer before merging this then that would be great!
> 
> Code looks 👍

Have spoken to Zef and altered the padding slightly. Will add the comment about the journey param now!

PR #525, opened 2019-03-29
Update journeys for payment failure text

laurenemms:
Following a previous PR that added some additional text to the sign in pages for uses in the payment failure journey https://github.com/guardian/identity-frontend/pull/524, this PR creates a clientId for journey.

The reason for this is that the payment email is triggered by
1. payment failure
2. issues with direct debit
3. credit card expiry

This also handles the messaging for each journey. This new messaging has been checked by Jess.

Payment Failure
Screen Shot 2019-03-29 at 17 03 49

Direct Debit Issue
Screen Shot 2019-03-29 at 17 04 15

Credit Card Expiry
Screen Shot 2019-03-29 at 17 04 01


app/com/gu/identity/frontend/views/models/TwoStepSignInStartViewModel.scala
```diff
@@ -94,7 +94,11 @@ object TwoStepSignInStartViewModel {
       resources = resources,
       indirectResources = layout.indirectResources,
       actions = actions,
-      paymentFailure = journey.contains("paymentFailure")
+      journey = Map(
+        ("paymentFailure", journey.contains(PaymentFailureJourney)),
```
jfsoul: Did you think about making this a single value rather than a `Map[String, Boolean]`?  Maybe it needs to be a bit like this for handlebars templates.

laurenemms: Yess, the reason I did it like this is because of the limited functionality existing in handlebars. I couldn't seem to find any functionality where I could do "if journey == "paymentFailure". The only if statements you seem to be able to use have to be like "#if journey", which should return true or false. Alternatively, I could model the view model as ``paymentFailiure = Boolean,
      creditCardExpiry = Boolean,
      directDebitIssue = Boolean
`` then set each using journey.contains but I thought this would be cleaner

jfsoul: I wondered whether it would be instead be possible to pass the message to the template, so it becomes `journeyMessage = String`.  I actually don't think that idea is particularly extensible though, because it won't expand nicely to other things we might want to change about the page depending on the journey.

app/com/gu/identity/frontend/views/models/TwoStepSignInStartViewModel.scala
```diff
@@ -94,7 +94,11 @@ object TwoStepSignInStartViewModel {
...
+        ("directDebitIssue", journey.contains(DirectDebitIssue)),
+        ("creditCardExpiry", journey.contains(CreditCardExpiry))
```
jfsoul: can this be `equals` rather than `contains`?

laurenemms: I just tried to but IntelliJ is throwing a warning because it doesn't like the type difference between journey (Option[Journey]) and PaymentFailureJourney.type

PR #532, opened 2019-05-15
SC_GU_RS cookie users to second step of sign in: User Details Endpoint 

paddyo:
# [Take users to second step of sign in after leaving scope of cookie](https://trello.com/c/73f3O00c/1371-take-users-to-second-step-of-sign-in-after-leaving-scope-of-cookie)

## Description
This is the second part of the update to use the GC_GU_RS cookie to allow the automatic filling in of the email address on the second part of the sign in.

## Changes
* Add a `getUserDetails` function to `IdentityClient` that can echo a provided cookie the `user details endpoint`
* Add a `getUserDatails` function to `IdentityService` to expose the equivalent `IdentityClient` functionality.
* Add a `UserDetailsResponse` for representation and deserialisation of a `/user/details` api call.
* Add a `SignInWithRSCookie` path in the Application controller for using the GC_GU_RS cookie to derive the email address.
* Refactor Application methods notably `firstStepSignIn` to deduplicate code.
* Add `MissingCookie` application exception for cases where a required cookie is not found.
 
## Depends on
* guardian/identity#1537


app/com/gu/identity/frontend/services/IdentityService.scala
```diff
@@ -52,6 +52,7 @@ trait IdentityService {
   def decryptEmail(encryptedEmail: String)(implicit ec: ExecutionContext): EitherT[Future, AppException, String]
   def getUserSignInDetails(email: String)(implicit ec: ExecutionContext): EitherT[Future, AppException, UserSignInDetails]
   def useAutoSignInToken(token: String)(implicit ec: ExecutionContext): EitherT[Future, AppException, PlayCookies]
+  def getUserDetails(cookie: PlayCookie)(implicit ec: ExecutionContext): EitherT[Future, AppException, String]
```
Mullefa: At the service layer this method could be called `getUserEmailAddress()` ?

app/com/gu/identity/service/client/ApiRequests.scala
```diff
@@ -38,6 +39,17 @@ object ApiRequest {
     apiEndpoint(path)
   }
 
+  def convertToWSCookie(cookie: Cookie): DefaultWSCookie =
```
Mullefa: Maybe a comment justifying this method i.e. _the docs suggested... but actually..._ Then at least it flags up the potential to simplify in the future (to devs working on the code base).

app/com/gu/identity/frontend/services/IdentityService.scala
```diff
@@ -288,4 +289,18 @@ class IdentityServiceImpl(config: Configuration, adapter: IdentityServiceRequest
...
   }
+
+  override def getUserDetails(cookie: PlayCookie)(implicit ec: ExecutionContext): EitherT[Future, AppException, String] = {
+    client.getUserDetails(cookie)
+      .bimap(
+        errors => {
+          logger.error(s"unable to get user details via cookie - errors: $errors")
+          appExceptionFromIdentityClientErrors(errors)
+        },
+        response => {
```
Mullefa: syntax: can just one line this function

app/com/gu/identity/service/client/ApiResponses.scala
```diff
@@ -88,3 +88,8 @@ case class UserSignInDetailsResponse(userSignInDetails: UserSignInDetails)
 object UserSignInDetailsResponse {
   implicit val userSignInDetailsResponseReads: Reads[UserSignInDetailsResponse] = Json.reads[UserSignInDetailsResponse]
 }
+case class UserDetailsResponse(email: String)
+
+object UserDetailsResponse {
+  implicit val userDetailsResponse: Reads[UserDetailsResponse] = Json.reads[UserDetailsResponse]
```
Mullefa: Think `userDetailsResponseRead` is preferable - scoping the variable name to the type class means that type class instances don't clash (in event of us having to put other type class instances in the companion object).

app/com/gu/identity/frontend/controllers/Application.scala
```diff
@@ -102,10 +101,20 @@ class Application(
         .discardingCookies(DiscardingCookie("GU_SIGNIN_EMAIL", "/", Some(configuration.identityCookieDomain)))
     }
 
+  def signInWithRSCookie(signInData: SignInRequestData)(implicit messages: Messages): EitherT[Future, AppException, Result] =
+    for {
+      cookie <- EitherT.fromOption[Future](signInData.request.cookies.get("SC_GU_RS"), MissingCookie("SC_GU_RS"))
+      emailOpt <- identityService.getUserEmailAddress(cookie)
+      email <- EitherT.fromOption[Future](emailOpt, MissingCookie("SC_GU_RS"): AppException)
```
Mullefa: Not sure if we want to return a `MissingCookie` exception at this point. One thing we could do is update the return signature of `IdentityService#getUserEmailAddress` to be of type `EitherT[Future, AppException, String]` and handle the case of an undefined email address in this method.

PR #534, opened 2019-05-29
Update core dependencies

mchv:
Update core dependencies to their latest stable when compatible:
- sbt and associated plugins
- scala
- akka
- playfamework

It need to be tested at least locally before merging


Mullefa:
Thanks Mariot, we'll test on CODE, then merge.

Mullefa:
@guardian/identity tested on CODE - will merge tomorrow.

PR #539, opened 2019-06-11
Ask user for password if they're blocked from social sign-in

Mullefa:
Adds functionality so that if a user is blocked from social sign-in, they will get a message explaining this, and be prompted to enter their password instead.

blocked_from_social_sign_in


app/com/gu/identity/frontend/controllers/Redirects.scala
```diff
@@ -22,21 +23,30 @@ final class Redirects(cc: ControllerComponents)
...
+  // (see corresponding comment in routes file for more information).
+  // However, keep this action for now to facilitate local integrated development with identity-federation-api.
+  // Forward query parameters to replicate behaviour in CODE and PROD.
   def registerPageTrailingSlash =
-    redirectToRoute(registerRoute, isLegacy = true)
+    redirectToRoute(registerRoute, isLegacy = true, forwardQueryParams = true)
+
+  private def redirectToRoute(call: Call, isLegacy: Boolean = false, forwardQueryParams: Boolean = false) = Action { request =>
+    val url =
+      if (forwardQueryParams) {
```
dominickendrick: Could this be pulled out as a separate function and tested ? Just so it is clear we are going from a `Map["param1" -> Seq("value1","value2")]` to a `Seq(("param1","value1"),("param1","value2"))`

app/com/gu/identity/frontend/errors/ErrorID.scala
```diff
@@ -79,7 +79,11 @@ object ErrorIDs {
     val key = "fbEmail"
   }
 
-
+  // Error code returned by Federation API if a user needs to verify their email before signing in with social sign-in.
+  // See identity-federation-api repo for more context / details.
+  case object AccountLinkingRequired extends ErrorID {
+    val key = "accountLinkingRequired"
```
dominickendrick: Could the format of the key be `"account-linking-required"` like the one below ?

app/com/gu/identity/frontend/controllers/Application.scala
```diff
@@ -221,15 +259,24 @@ class Application(
...
+        // try and sign-in via social sign-in, fail because it was blocked,
+        // and then go through the RS cookie sign-in flow again,
+        // rather than being notified social sign-in is blocked for them.
+        .recoverWith { case err =>
+          if (!isInapplicableSignInFlow(err)) logger.error("unable to sign-in with email token")
+          signInWithPasswordIfUserBlockedFromSocialSignIn(signInData)
+        }
         .recoverWith { case err =>
-          if (!isInapplicableSignInFlow(err)) logger.error("unable to sign-in with email token", err)
+          if (!isInapplicableSignInFlow(err)) logger.error("unable to sign-in via 'user blocked from social sign-in' flow", err)
```
dominickendrick: I know this was done in a prev PR but is `isBlockedSignInFlow` more descriptive ?

app/com/gu/identity/frontend/controllers/Redirects.scala
Mullefa: sure

app/com/gu/identity/frontend/controllers/Application.scala
```diff
@@ -151,6 +158,37 @@ class Application(
...
+  def signInWithPasswordIfUserBlockedFromSocialSignIn(signInData: SignInRequestData): EitherT[Future, AppException, Result] = {
+    EitherT.fromEither[Future]{
+      for {
+        _ <- Either.fromOption(signInData.errors.find(_ == AccountLinkingRequired.key), InapplicableSignInFlow("user not blocked from social sign-in"))
+        email <- Either.fromOption(signInData.email, EmailNotFound("user blocked from social sign-in, but email address not found"))
+        signInEmailCookie = CookieService.signInEmailCookie(email)(configuration)
+      } yield {
+        Redirect(
+          UrlBuilder(
+            routes.Application.twoStepSignInChoices(
```
dominickendrick: This is slightly unclear what Url is being returned here ? Could this be defined as a `val reEnterPasswordUrl = ...` for a clearer description ?

app/com/gu/identity/frontend/errors/ErrorID.scala
Mullefa: The current errors codes returned by identity-federation-api are in `camelCase` e.g. `fbEmail` (above this error), so I kept format the same for consistency.

app/com/gu/identity/frontend/controllers/Application.scala
Mullefa: `InapplicableSignInFlow` is just used to encode any errors that we don't care about in the context of the sign-in flow (#538) i.e. there was an error because the user wasn't applicable for the given sign-in flow (rather than they were, but e.g. an IO error occurred when making a request to the identity API). All of these checks are saying "only log it if it's an error we care about / didn't expect". When put like this, maybe using the function: 

```scala
def isUnexpectedError(err: AppException): Boolean = !isInapplicableSignInFlow(err)
```

would be clearer? I'd be wary of using something like `isBlockedSignInFlow()` because the whole point of (#538) was to encode all errors that we don't care about using one case class (so we can determine which errors to safely ignore without knowing the implementation of a given sign-in flow).

dominickendrick: Ah ok, I understand now. `UserEligibleSignInFlow` was discussed IRL

PR #550, opened 2019-08-08
Rename the temporary endpoint(s) for Sign in With Apple

lmath:
https://github.com/guardian/identity-frontend/pull/549 added the sign in with apple button accessible on CODE only via hitting `/demo` to be able to get auth codes for testing

but even though this is temporary, that's a not a great name

this renames demo => apple sign in button 

note: this depends on these changes being live in the code profile fastly vcl: https://github.com/guardian/identity-platform/pull/339


public/apple-sign-in-button.hbs
```diff
@@ -6,10 +6,10 @@
 <div id="appleid-signin" data-color="black" data-border="true" data-type="sign in"></div>
 <script type="text/javascript">
   AppleID.auth.init({
-    clientId : 'com.dev-theguardian.code.idapi',
+    clientId : 'com.dev-theguardian.code.profile',
```
lmath: We have created a new service id configuration in the apple developer console that has the redirect uri specified in this request

app/com/gu/identity/frontend/controllers/AppleSignInButtonController.scala
```diff
@@ -13,19 +13,19 @@ import scala.concurrent.ExecutionContext
 //
```
vlbee: Comment still says /demo

PR #553, opened 2019-09-24
Use new authentication model

Mullefa:
Use `identity-auth-play` instead of `identity-cookie` (implements the changes in authentication outlined [here](https://github.com/guardian/identity-platform/blob/master/docs/adr/0001-authentication-requires-invalid-sessions-check.md)).


app/com/gu/identity/frontend/authentication/UserAuthenticatedAction.scala
```diff
@@ -2,54 +2,114 @@ package com.gu.identity.frontend.authentication
...
+      // An alternative would have been to just use the cookie value (String)
+      // instead of a cookie, but this would have required significant re-factorisation
+      // of the identity-frontend application.
+      cookie <- getSCGUUCookie(request)
+    } yield {
+      cookie
+    }
+  }
+
+  def refine[A](request: Request[A]): Future[Either[Result, UserAuthenticatedRequest[A]]] = {
```
dominickendrick: Should we return a `EitherT[Future, Result, Double, UserAuthenticatedRequest[A]]` ?

Mullefa: `refine` is getting over ridden (implemented); it's defined in `ActionRefiner`; therefore, return type is set as is. I've added an override modifier to make this clearer.

dominickendrick: 👍

PR #560, opened 2019-10-30
Remove Sentry from serverside code

vlbee:
Removing sentry from logging serverside errors in Identity Frontend, as this tool was no longer being actively used or monitored by the team and the raven library used to implement has since been deprecated. 

TODO 
- [ ] Update config files after merge.


vlbee:
Tabling this for the time being

PR #561, opened 2019-10-30
Bump jackson-databind to 2.9.10.1

lmath:
Snyk flagged a vulnerability relating to the version of `jackson-databind` in `identity-frontend` https://app.snyk.io/vuln/SNYK-JAVA-COMFASTERXMLJACKSONCORE-471943

I inspected the dependency graph to look for outdated dependencies that pull in jackson-databind to see if bumping versions would pull in a sufficiently new version of `jackson-databind` without me needing to explicitly pull in `jackson-databind` but no dice.

So, this PR explicitly pulls in `jackson-databind`.

### To check in CODE before merge
- [x] Sign in - new user
- [x] Sign in - existing user
- [x] Reset password


build.sbt
```diff
@@ -17,8 +17,8 @@ resolvers += "Guardian Github Releases" at "https://guardian.github.io/maven/rep
 
 val identityLibrariesVersion = "3.181"
 val akkaVersion = "2.5.22"
-val playJsonVersion = "2.6.12"
-val awsSdkVersion = "1.11.503"
+val playJsonVersion = "2.7.4"
+val awsSdkVersion = "1.11.661"
```
lmath: These depend on jackson-databind so I tried bumping them first and then thought why not use new versions even if they don't help with this vulnerability

PR #564, opened 2019-11-04
Fix Javascript vulnerabilities

vlbee:
This PR takes our Snyk vulnerabilities down from this: 
image

to this :
image

### Steps: 

1. Ran `npm audit fix --package-lock-only` cli command.

This updated the dependency graph packages to the the minimum version for an accepted major range. Ie. `7.3.1 --> ^7.0.0` so all patches and minor releases will be updatable easily. 

The very manageable `snyk test` report at this stage: 
image

2. Manually updated the packages below:
* eslint
* eslint-plugin-import
* optimize-css-assets-webpack-plugin
* postcss-cli
* concurrently

### Tested in CODE:
- [x] Signin new user
- [x] Signin existing user
- [x] password reset

### Moral of the story

Don't lock your packages to a specific version without having a very good reason to

### Outstanding vulnerability

A webpack medium severity issue that breaks everything when I try to change it because webpack.... leaving for another PR


vlbee:
I think this can fully replace the Snyk patch PR #563

PR #566, opened 2019-11-11
Add ability to pass Ophan ComponentEvent data as a query parameter

coldlink:
As part of the Sign In Gate testing process, we want to be able to track user's that go on to Sign In or Create a Guardian account from our sign in gate component. While the ability to track events using Ophan existed, the ability to see if a user went on to sign in did not.

To do this we decided to use Ophan ComponentEvent tracking firing from Identity API (IDAPI), however we needed a way to pass information that the ComponentEvent requires through identity-frontend to IDAPI so that we could track individual users that went on to sign in.

We thought that the best way to do this would be to add on the parameters as part of the `Tracking Data` https://github.com/guardian/identity-frontend/blob/3922324c7cfbfedd421300713ed7036c744efbe5/app/com/gu/identity/frontend/models/TrackingData.scala#L7 that is sent to IDAPI as part of Sign In and Register actions, which we called `componentEventParams`.

To get the `componentEventParams` to be included in the tracking data, we needed to be able to pass on all the information all the way from guardian/frontend through the sign in and register flows in guardian/identity-frontend. We decided that the simplest way to achieve this would be to use URL query parameter, and pass on the information through that.

However due to the number of keys required to fire a component event (7), it wasn't feasible to pass each of these as separate query parameters, as we found it difficult to manage where, when and how query parameters were being passed through the login flow, as well as in some cases hitting the Scala max parameter limit of 22.

We decided that to get around this we would create a query string with all the parameters required, e.g. `componentType=SIGN_IN_GATE&componentId=inital_test&abTestName=SignInGateFirstTest&abTestVariant=variant&viewId=aBc123XyZ987&browserId=abc123&visitId=xyz789` then URL encode these, so the string would become `componentType%3DSIGN_IN_GATE%26componentId%3Dinital_test%26abTestName%3DSignInGateFirstTest%26abTestVariant%3Dvariant%26viewId%3DaBc123XyZ987%26browserId%3Dabc123%26visitId%3Dxyz789` 

and then assign this string to a key of `componentEventParams`, 

so the query parameter which we would attach to url/flow would be `componentEventParams=componentType%3DSIGN_IN_GATE%26componentId%3Dinital_test%26abTestName%3DSignInGateFirstTest%26abTestVariant%3Dvariant%26viewId%3DaBc123XyZ987%26browserId%3Dabc123%26visitId%3Dxyz789`. 

This meant that we would only have to manage 1 parameter in the identity-frontend flow compared to 7. Also since we encode this in guardian/frontend and decode this in IDAPI, identity-frontend doesn't have to deal with the implementation of this, hence we can just pass it around as an `Option[String]`.

We now just include `componentEventParams` as a query parameter that routes in identity-frontend use, and pass around this the say way as other query parameters, until we can include this in the tracking data that is sent to IDAPI.

This PR also included the UrlBuilder service changes made in (https://github.com/guardian/identity-frontend/pull/567) but that should be merged in first before this.

- [x] Add PR descriptions
- [x] Add inline comments for what `componentEventParams` is and how it is used
- [x] Should we validate the input before we send it to IDAPI


app/com/gu/identity/frontend/controllers/SigninAction.scala
```diff
@@ -148,20 +152,29 @@ class SigninAction(
...
       skipConfirmation = None,
-      skipValidationReturn = None
+      skipValidationReturn = None,
+      componentEventParams = None
     )
 
   def emailSignInFirstStep: Action[SignInActionRequestBody] = SignInServiceAction(bodyParser) { implicit request: Request[SignInActionRequestBody] =>
     identityService.getUserType(request.body.email)
       .map { userType =>
+
```
dominickendrick: Can we add some comments describing why we are adding these extra parameters

app/com/gu/identity/frontend/models/TrackingData.scala
```diff
@@ -9,27 +9,43 @@ case class TrackingData(returnUrl:Option[String],
...
-    userAgent.map("trackingUserAgent" -> _)
+    userAgent.map("trackingUserAgent" -> _),
+    componentEventParams.map("componentEventParams" -> _)
   ).flatten
 }
 
 object TrackingData extends RemoteAddress {
 
-  def apply(request: RequestHeader, returnUrl: Option[String]): TrackingData = {
+  def apply(request: RequestHeader, returnUrl: Option[String], componentEventParams: Option[String]): TrackingData = {
```
dominickendrick: We can just add a default value for `componentEventParams: Option[String] = None` and remove the `apply` below

PR #567, opened 2019-11-12
Cleanup UrlBuilder Service

coldlink:
While implementing tracking for the Sign In Gate component, we discovered that `UrlBuilder` service, a service which appended query parameters to a given URL (or path), was a mess with many apply functions for every possible combination of parameters it required calling other apply methods, this made it difficult to read, utilise, or extend it's functionality. 

We abstracted these apply methods into `buildUrlWithFlowParameters` and `mergeUrlAndParameters` methods.

The `mergeUrlAndParameters` takes a `baseUrl` and `params`, it converts the params from a `Seq[(String, String)]` to the query-string format, and appends this onto the `baseUrl`.

The `buildUrlWithFlowParameters` method takes a `baseUrl` and any of the named parameters with defaults. This means that instead of having multiple apply methods for the parameter that we wanted, we can now just call this new methods, and pass named parameters (e.g. `UrlBuilder.buildUrlWithFlowParameters(baseUrl, returnUrl = Some(finalReturnUrl), skipConfirmation = skipConfirmation, clientId = clientId)` for the parameters that we need.

A final apply method took the Play representation of queries `Map[String, Seq[String]]` and converts that to a query string too, which we've now called `appendPlayQueryStringToPath`.


app/com/gu/identity/frontend/models/UrlBuilder.scala
```diff
@@ -2,79 +2,61 @@ package com.gu.identity.frontend.models
...
-
-  def apply(baseUrl: String, call: Call): String = s"$baseUrl${call.url}"
-
-  def apply(call: Call, intcmp: Option[String]): String =
-    apply(call.url, buildParams(intcmp = intcmp))
-
-  def apply(call: Call, intcmp: Option[String], abTest: Option[AbTest]): String =
-    apply(call.url, buildParams(intcmp = intcmp, abTest = abTest))
+  // takes a base url (or path) and any required query parameters (be sure to use named parameters so you can pass only
+  // the parameters that are required), and converts and appends these parameters to the base url
```
dominickendrick: Could we also add that ` all these parameters are used within the signin and registration flows for tracking and routing the user, depending on their state or the flow required`

coldlink: 👍 Made the change

coldlink: Made the change 👍

vlbee:
thanks for doing this. Way clearer!

PR #571, opened 2019-12-23
Change desired instances in PROD from 3 to 5

lmath:
We'd like to change our desired number of instances in PROD from 3 to 5 over just for the holiday period.


lmath:
Tested changing number of desired instances in CODE by deploying this build to CODE: https://github.com/guardian/identity-frontend/pull/571/commits/6bd9141941bf41878a6ace91dc74fc93eb803d41 and saw 2 instances of identity frontend code with max 4 instances in the CODE ASG.

cloudformation/identity-frontend.yaml
```diff
@@ -64,8 +64,8 @@ Mappings:
       LatencyAlarmPeriod: 60
       NotificationAlarmPeriod: 1200
       InstanceType: t3.small
-      DesiredInstances: 3
-      MaxInstances: 12
+      DesiredInstances: 5
+      MaxInstances: 20
```
lmath: I'm not totally sure why there is a desired instances * 4 = max instances relationship, but I'll maintain that for now during this temporary bump from 3->5 desired instances.

jacobwinch: @lmath - I would guess that it's to provide some headroom in case you need to scale up automatically: https://github.com/guardian/identity-frontend/blob/master/cloudformation/identity-frontend.yaml#L346-L373

adamnfish: larger than 2x is useful to allow deployments while the application has scaled up due to traffic.

So given we're looking for >2x, why 4x?
4x is a significant number because if a deploy fails spuriously you can hit deploy again straight away and it'll fix itself. Of course, this is only true if the app isn't scaled up.

adamnfish: p.s. It's not clear to me from this fragment what the minimum is set to. Typically the desired starts at the minimum and we let traffic / deployments fiddle with it from there. We often omit the desired from the CloudFormation template so that a CloudFormation update doesn't reset the Autoscaling Group's size if it's in the middle of scaling up or down for whatever reason (e.g. as above traffic load / deployment).

Nothing to worry about for now, just an observation!

pvighi: If it is normal for this to scale out to more instances it might be to still allow deployment during those times. I think you need to have max instances set to at least double your current instance count to be able to deploy using riff-raff.
But I might be completely wrong as I never worked on this so this is complete guess

lmath: thanks @jacobwinch - TIL identity-frontend scales up automatically 😳 
thanks @adamnfish and @pvighi  - those explanations are helpful, and @adamnfish when we go back to reverse the change in this PR that seems like a good time to tidy up and specify min instances instead

PR #584, opened 2020-05-19
Dynamic Recaptcha for identity

QuarpT:
#### Work done

Implement recaptcha for Identity frontend password login. If Google thinks the user might be a bot, the captcha is displayed  when the user clicks sign in. On completing the captcha successfully, the login request is posted.

Recaptcha is only displayed if one of the alarms for Login failure goes off.

There are 3 alarms,
1) 500 failed logins in one minute in the past an hour
2) 300 failed logins in one minute 5 times in the past hour
3) 150 failed logins in one minute 20 times in the past hour

The alarm states are polled on five minute intervals past the hour to keep the instances serving the same captcha switch state.

On captcha failure the login page refreshes with the captcha rendered. The solves the issues arising when the instances are in different captcha switch states because of a race condition, or if the captcha switch turned on since the user loaded a login page.

Chrome smartlock will fail when captcha is applied, but since the captcha is temporary it's not a huge issue.

In JavaScript, the recaptcha can be initialized 
- on ajax step flow as an identity frontend js component, if the user transitions from the enter email form
- on the www.google.com/recaptcha/api.js loaded event, when the login page is refreshed (it's loaded in the <head> as a <script> since there is no npm package for recaptcha)

Removed:

- Partially implemented captcha which was mostly removed in 2016
- Remove unused login page

#### Why are we doing this:

In the past bots have brute forced the password login endpoint while avoiding the rate limiter by attacking from a large set of IPs against a large set of email addresses. A captcha should prevent bots receiving password correct/incorrect responses. 

The captcha is only displayed during an attack. This means users without JavaScript can continue to authenticate most of the time

Screenshot 2020-05-19 at 16 40 45

Screenshot 2020-05-19 at 16 41 34

Screenshot 2020-05-19 at 16 44 26


app/com/gu/identity/frontend/controllers/SigninAction.scala
```diff
@@ -91,10 +93,6 @@ class SigninAction(
     signInAction(successfulSignInResponse, successfulAjaxSignInResponse, signInMetricsLogger)
   }
 
-  def signIn = SignInServiceAction(bodyParser) {
```
QuarpT, line 94: this signin endpoint does not appear to be called

app/com/gu/identity/frontend/recaptcha/GoogleRecaptchaCheck.scala
```diff
@@ -1,27 +0,0 @@
-package com.gu.identity.frontend.controllers
-
-import com.gu.identity.frontend.services.GoogleRecaptchaServiceHandler
-import play.api.mvc.Result
-
-import scala.concurrent.{ExecutionContext, Future}
-
-class GoogleRecaptchaCheck(googleRecaptchaServiceHandler: GoogleRecaptchaServiceHandler) {
```
QuarpT, line 8: old captcha code that was left behind in 2016

app/com/gu/identity/frontend/recaptcha/GoogleRecaptchaServiceHandler.scala
```diff
@@ -1,57 +0,0 @@
-package com.gu.identity.frontend.services
```
QuarpT, line 1: old recaptcha code from 2016

conf/CODE.conf
```diff
@@ -13,8 +13,6 @@ identity.frontend.dotcomBaseUrl="http://m.code.dev-theguardian.com"
 
 identity.federation.api.host="https://oauth.code.dev-theguardian.com"
 
-google.recaptchaEnabled = false
```
QuarpT, line 16: these are flags from the implementation removed in 2016

public/components/ajax-step-flow/ajax-step-flow__slide.js
```diff
@@ -174,10 +181,18 @@ const init = ($slide: HTMLElement): void => {
...
-      });
+      displayCaptcha($form)
+        .then(() => {
+          fetchAndDispatchSlide($form.action, $slide, {
+            method: 'post',
+            body: new FormData($form)
+          });
+        })
+        .catch(error => {
+          if (error !== captchaResubmitted) {
```
QuarpT, line 192: the promise fails if the user starts a new captcha

app/com/gu/identity/frontend/recaptcha/CaptchaSwitchService.scala
```diff
@@ -0,0 +1,72 @@
...
+import java.util.concurrent.atomic.AtomicBoolean
+import java.util.concurrent.{Executors, TimeUnit}
+
+import com.amazonaws.services.cloudwatch.AmazonCloudWatchAsync
+import com.amazonaws.services.cloudwatch.model.{DescribeAlarmsRequest, DescribeAlarmsResult}
+import com.gu.identity.frontend.logging.{CloudWatchClient, Logging}
+
+import scala.collection.JavaConverters._
+import scala.concurrent.duration._
+
```
dominickendrick, line 14: Could we add a comment that describes why we are polling alarms etc, I think the PR description is a good candidate - we aim to follow the principles layed out in this blog http://antirez.com/news/124 post around Why and Design comments to make it easier for the future developer to know why its been implemented in the current way

app/com/gu/identity/frontend/recaptcha/CaptchaSwitchService.scala
```diff
@@ -0,0 +1,72 @@
...
+  private def scheduleUpdateCaptchaSwitch(): Unit = {
+    val runnable: Runnable = () => updateCaptchaSwitch()
+    scheduler.schedule(runnable, timeUntilFiveMinutes(Instant.now), TimeUnit.MILLISECONDS)
+  }
+
+  private def updateCaptchaSwitch(): Unit = try {
+    logger.info("updateCaptchaSwitchBasedOnAlarms execute")
+
+    val describeAlarmsRequest = new DescribeAlarmsRequest().withAlarmNames(
+      s"CaptchaOnLoginFailure1${CloudWatchClient.stageDimension}",
```
dominickendrick, line 34: I think its great we are polling the alarms, it would be good to understand how we can validate that the threshold for displaying captcha is correct 🤔

app/com/gu/identity/frontend/recaptcha/CaptchaSwitchService.scala
```diff
@@ -0,0 +1,72 @@
...
+import com.gu.identity.frontend.logging.{CloudWatchClient, Logging}
+
+import scala.collection.JavaConverters._
+import scala.concurrent.duration._
+
+class CaptchaSwitchService(cloudWatchClient: AmazonCloudWatchAsync) extends Logging {
+  import CaptchaSwitchService._
+
+  private val scheduler = Executors.newScheduledThreadPool(1)
+  private val captchaSwitch = new AtomicBoolean(false)
```
dominickendrick, line 19: 👏 nice use of `AtomicBoolean` and a good example of a scheduled service

app/com/gu/identity/frontend/recaptcha/CaptchaSwitchService.scala
QuarpT: 👍 I will add some information here

app/com/gu/identity/frontend/recaptcha/CaptchaSwitchService.scala
QuarpT: My original plan was to implement a sliding window for intervals of 1 minute, 5 minutes and 60 minutes, reading from the signin failure metric manually (sliding windows aren't possible with alarms)

The thresholds were set to 20x the maximum normal signin failure witnessed in the last 2 weeks in production for those periods.

I decided to change the implementation to use AWS alarms instead, which are easier to maintain. AWS alarms do not support sliding windows, but does support ['M of N' alarms](https://aws.amazon.com/about-aws/whats-new/2017/12/amazon-cloudwatch-alarms-now-alerts-you-when-any-m-out-of-n-metric-datapoints-in-an-interval-are-above-your-threshold/)

I tried to approximate the sliding window thresholds with 'M of N' alarms. The approximation is bit arbitrary, so it can be debated 😆

QuarpT: I chose not to use a sigin failure/success ratio, since a bot can make successful sign in attempts for an account it owns to control the ratio.

dominickendrick: Fair point, lets see how it pans out in production, and we can change the threshold if needed : )

PR #592, opened 2020-06-09
Registration captcha

QuarpT:
## What does this change?

- Add Google Recaptcha on user registration with password
- Remove captcha switch service and triggering alarms

## Why are we doing this?

- We want a captcha on user registration to prevent bots making accounts and potential account squatting
- We are going to trial having the captcha being on all the time and check whether this affects the signin/registration rate.

## Question

Do we want to experiment with some users viewing the captcha, and some not?


public/components/register-form/_register-form.hbs
```diff
@@ -1,51 +0,0 @@
-<form id="register_form" class="register-form" action="{{ actions.register }}" method="POST">
```
QuarpT, line 1: unused

public/components/two-step-signin/two-step-signin-new.hbs
```diff
@@ -15,6 +15,8 @@
       <input class="form-input form-field-wrap__field" type="email" name="email" id="tsse-email" autocomplete="username" autocapitalize="off" autocorrect="off" spellcheck="false" readonly value="{{email}}" />
     </div>
 
+    <div class="sign-in-recaptcha"></div>
```
QuarpT, line 18: this adds the captcha to the register form

PR #596, opened 2020-07-08
Propagate tracking to idapi

QuarpT:
## What does this change?

Propagate tracking information for the profile signin page to IDAPI. 
IDAPI uses these additional parameters to trigger a second Ophan event for the signin page view event.

We get a user ID and session associated with the profile.theguardian.com/signin page view.

## What is the cost of failure?

Users cannot login.


public/components/form/form-eventparams.js
```diff
@@ -0,0 +1,65 @@
...
+import { getCookie } from '../browser/cookie';
+
+const selector: string = '.componentEventParams';
+
+const appendEventParameter = (
+  element: Element,
+  parameterKey: string,
+  parameterValue: string
+) => {
+  const inputFormElement: Object = element;
```
QuarpT, line 12: flow complains about inputFormElement not having the value field unless we do this

PR #597, opened 2020-07-10
propagate tracking to federation

QuarpT:
## What does this change?

Propagate social link tracking parameters to federation api. The current page view is only accessible from JavaScript, so we add the URL parameters in JS, rather than server side.

All social link tracking parameters get sent to federation api where we extract them and send event data to ophan.


public/components/form/form-eventparams.js
```diff
@@ -4,6 +4,41 @@ import { getCookie } from '../browser/cookie';
 
 const selector: string = '.componentEventParams';
 
+const appendEventParamsToHrefOnce = (
+  socialSigninLink: Object,
+  eventParameters
+) => {
+  if (
+    !socialSigninLink.href ||
+    socialSigninLink.href.indexOf('componentEventParams=') !== -1
```
coldlink, line 13: this works, but i prefer using `includes` these days https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes)

QuarpT: `includes` has no IE support but I guess `includes` gets polyfilled

coldlink: yeah, using babel it should be transpiled down to ES5 by default.

PR #601, opened 2020-07-27
Send component event parameters for smartlock login from profile page

QuarpT:
## What does this change?

- Send component event parameters for smartlock login from profile page.
- Init components after Ophan is loaded since some components rely on Ophan pageview ID
- /reauthenticate also takes component event parameters (the same as /signin), so we can track users being reauthenticated by IDAPI

## How to test

Check that the componentEvent params get sent on profile page smartlock authentication

We are able to track profile page signin on Identity API


public/components/form/form-eventparams.js
```diff
@@ -85,18 +85,8 @@ const appendEventParametersToElement = (element: Element) => {
   }
 };
 
-const appendEventParametersToInputs = () => {
```
QuarpT, line 88: we don't need `appendEventParametersToInputs` since init is always called after Ophan is initialised

